PR #11, opened 2021-03-11
Database task simple implementation

theo-damiani:
Add a simple interface for the database to manage tasks.
And creation of a class Task.


app/src/main/java/com/github/steroidteam/todolist/todo/Task.java
```diff
@@ -0,0 +1,64 @@
...
+
+        this.id = id;
+        this.body = body;
+    }
+
+    /**
+     * Modifies the body of the task.
+     * @param newBody The new body.
+     */
+    public void modifyBody(String newBody) {
```
YagoGG, line 33: Following Java's convention of getters and setters, I feel like `setBody` is a better name here. What do you think?

app/src/main/java/com/github/steroidteam/todolist/todo/Task.java
```diff
@@ -0,0 +1,64 @@
...
+    private final String id;
+
+    /**
+     * Constructs a new Task with a title and a body.
+     *
+     * @param id The id of the task.
+     * @param body The body of the task.
+     * @throws IllegalArgumentException Thrown if one or more arguments are null
+     */
+    public Task(String id, String body) {
```
YagoGG, line 20: I am wondering if it makes sense to keep the ID in each `Task` object, or at the very least if we can maybe provide the ID afterwards.

My concern is that the ID is usually task-agnostic: it mostly depends on the database and the elements it already contains. We probably will not know the ID assigned to this task until we try to store it in the database, and its get assigned an ID.

This is problematic, because that means that we would not be able to do things like this:

```java
// Create "database"...

Task myTask = new Task("Do something"); // No ID supplied.
database.addTask(myTask); // ID assigned here.
```

A possible fix I can think of is to allow `id` to be `null` upon creation, and then let the database assign it. Not sure if it is the optimal pattern though. Thoughts?

app/src/main/java/com/github/steroidteam/todolist/todo/Task.java
```diff
@@ -0,0 +1,64 @@
...
+
+import java.io.Serializable;
+
+/**
+ * A class that represents a task described by a title and a body.
+ */
+public class Task implements Serializable{
+
+    private String body;
+    private final String id;
```
YagoGG, line 11: I would go with an (unsigned?) int for the ID.

app/src/main/java/com/github/steroidteam/todolist/database/DatabaseException.java
```diff
@@ -0,0 +1,9 @@
+package com.github.steroidteam.todolist.database;
+
+public final class DatabaseException extends Exception{
```
YagoGG, line 3: Missing space after `Exception`.

YagoGG (changes requested):
Hey @theo-damiani! Thanks a lot for the PR.

I have left a few in-line comments, please take a look at them and let me know if you have any questions.

The CI is complaining because of the code coverage; apparently it is not recognizing your tests for some reason. Could you look into it?

Also, it would be great if you could add a `VolatileDatabase` that implements your `Database` and just keeps the contents in a variable. That way we can integrate your work with the actual logic.

Cheers! :)

app/src/main/java/com/github/steroidteam/todolist/todo/Task.java
theo-damiani: For sure, I will follow getters and setters convention.

app/src/main/java/com/github/steroidteam/todolist/todo/Task.java
theo-damiani: I go for an Integer, so we can create a task with null id:  not already in the database.

app/src/main/java/com/github/steroidteam/todolist/todo/Task.java
theo-damiani: I see what you mean.
I remove the Id from the constructors and so it is assigned by the database when there is a request to put the task into it.
I created an Id manager (for now very simple just an incrementation) that is responsible to assign a valid id to a task. 
So we can work on the database and the task more independently. We will see how goes the Id management with the authentication system.

app/src/main/java/com/github/steroidteam/todolist/database/DatabaseException.java
theo-damiani: Corrected.

YagoGG (approved):
Looks amazing, thanks for the great job! :tada:

shoutizix:
Seems great !

shoutizix (approved):

PR #13, opened 2021-03-11
Simple Controller Implementation

shoutizix:


app/src/main/java/com/github/steroidteam/todolist/MVC/AppController.java
```diff
@@ -0,0 +1,55 @@
...
+            throw new IllegalArgumentException();
+        }
+        try {
+            taskDatabaseAdapter.updateTask(new Task(bodyNewTask));
+        } catch (DatabaseException e) {
+            e.printStackTrace();
+        }
+    }
+    */
+    // TODO removeTask since it needs to know which id the view has assigned to the task I don't know how to do it for now
```
VPellenc, line 51: From how I see it, the view should have the whole Task object available, i.e. it's String content to display it, but also its id in order to modify/remove. If this is true this means the taskDatabaseAdapter should return the id/the task when we create a task.
Otherwise the AppController would have to keep a mapping between local view task id, and the corresponding id in the database.

Correct me if I missed something

VPellenc (approved):
Supporting the ability to create task in the controller will be helpful very soon.
The tests passes and show the correctness of the feature as far as I can see.

It can be questioned to merge this as the controller is not complete but I'm in favor since it will reduce the dependency needs for the next sprint.

sydneyhauke (approved):
Seems to be a good start for the AppController.

But we will definitely need to think more in depth about :
1) the different "services" this controller will provide. I expect that these "services" will be called by the different android activities of the app.
2) Which data, information, type of objects we will pass around in the provided methods. Indeed, how do we identify todoLists, a single todoList, a task etc... ?
3) Having the controller to be a SingletonObject. I don't think we will need more than one controller in the whole app.

So good start, but then we will have to improve this class further.

YagoGG:
Closing this as discussed today in the team standup, since it is going to be superseded by #76.

PR #16, opened 2021-03-11
Feature/layout

VPellenc:
Adds the ItemViewActivity, featuring its design. It does not have interactivity yet.


theo-damiani (approved):
Great job!
Nothing important needs to be changed for now. 
We will see, for example for the item list when we will include user interaction.

lgraziano (approved):

PR #19, opened 2021-03-11
logic: Implement in-memory task creation prototype

YagoGG:


VPellenc (approved):
Tested on my Android emulator and it works as expected.

theo-damiani (approved):

PR #36, opened 2021-03-12
Definition of the new Database Interface.

theo-damiani:
I re-define the Interface of the Database so that Yago can continue working on the firebase implementation.
I added a simple TodoList class.
And as we will no longer use adapters, I removed them.

Edit: Implementation of the VolatileDatabase and test for 100% branch coverage.


app/src/main/java/com/github/steroidteam/todolist/todo/Task.java
```diff
@@ -5,13 +5,13 @@
...
- * The id is null upon the creation of the Task, and it will
+ * The index is null upon the creation of the Task, and it will
  * be assigned by the database when it's pushed into it.
  */
-public class Task implements Serializable {
+public class Task {
 
     private String body;
-    private Integer id;
+    private Integer index;
```
VPellenc, line 15: Wouldn't it make sense to also add a boolean isChecked (or isDone) attribute, in order to remember which tasks the user has checked ?

YagoGG:
Hey @theo-damiani! Thank you very much for putting together this PR in so little time, it has been super useful (I am already using your commits in #37, in preparation for when this is merged).

I remember @shoutizix mentioning that this interface could be slightly hard to integrate with Android's view. I was wondering if using the observer pattern could be a good idea here, so that whatever `Database` we're using calls the observers that update the view whenever a task is created/modified/deleted.

Just to be clear, the `Database` interface would include methods to add & remove observers, and these observers could be event listeners for the view we're using for the todo list. This is assuming that the view supports event listeners, which is something we should confirm (I have not been able to do it myself yet).

Cheers!

app/src/main/java/com/github/steroidteam/todolist/todo/Task.java
theo-damiani: Sure, I will add it.

theo-damiani:
> 
> 
> Hey @theo-damiani! Thank you very much for putting together this PR in so little time, it has been super useful (I am already using your commits in #37, in preparation for when this is merged).
> 
> I remember @shoutizix mentioning that this interface could be slightly hard to integrate with Android's view. I was wondering if using the observer pattern could be a good idea here, so that whatever `Database` we're using calls the observers that update the view whenever a task is created/modified/deleted.
> 
> Just to be clear, the `Database` interface would include methods to add & remove observers, and these observers could be event listeners for the view we're using for the todo list. This is assuming that the view supports event listeners, which is something we should confirm (I have not been able to do it myself yet).
> 
> Cheers!

Thanks Yago !

I just saw with Noah what he needs to integrate the interface with the view. I will work on it.
I thought to use a BaseAdapter. So in a way, it is the observer and we just have to call the provided method notifyDataSetChanged();

Maybe we can merge this branch as it's called fix/... and I will create another one for the integration of the interface with the android view.

Cheers!

YagoGG:
Hey @theo-damiani!

After working in the implementation of the Firebase database, I noticed that some changes were needed in the `Database` interface and the `Task`/`TodoList` classes.

The modifications are minor, but I think it might be easier to review all of them in a new PR. It is exactly the same as your (preserving the authorship of your commits) but a few extra commits on top.

Would you be okay with superseding this PR with #41?

theo-damiani:
Hey @YagoGG,
Sure, no problem! 
Should I close this PR?

PR #45, opened 2021-03-17
db: Firebase-based filesystem service

YagoGG:
This PR adds the concept of a `FileStorageService`, as an API to interact with a filesystem (which can be local, or most likely remote).

We'll use this filesystem as the lower layer for our implementation of the database (e.g. whenever a `TodoList` is added, a new file is created).


theo-damiani (approved):
Awesome work!

I'm not really sure to fully understand, but it seems to be definitely great work.

I just have one question: 
Should you add some DatabaseException? for example, if there is an error when recovering a file. When you `completeExceptionally` just precise with an exception `completeExceptionally(new DatabaseException(""))`

YagoGG:
Hey @theo-damiani! Thank you very much for the review.

Yeah, even though I tried to make it as clean as possible, I know that it's hard to understand. Feel free to ping me if you feel like anything could be clearer or better documented (either now or in the future).

>  Should you add some DatabaseException? for example, if there is an error when recovering a file. When you `completeExceptionally` just precise with an exception `completeExceptionally(new DatabaseException(""))`

Keep in mind that the `FirebaseFileStorageService` does not implement our `Database`; it's a lower layer in the process. Thus it's completely agnostic of whether we're using it with a `Database` or something else.

The `DatabaseExceptions` will be raised by the 2nd PR I'm preparing, that takes care of adapting this `FirebaseFileStorageService` to a `FirebaseDatabase`.

Hope that makes it a bit clearer!

lgraziano (approved):
Nice ! I read all of it and it all seems fine. Plus, great work on the testing part !

PR #46, opened 2021-03-18
Feature/todo list controller

VPellenc:
I created this draft to get feedback on the comments I put in the code, and for the general structure of the controller.

Once everything is clear i'll clean up comments, add the tests, and make a proper PR.


app/src/main/java/com/github/steroidteam/todolist/MVC/AppController.java
```diff
@@ -0,0 +1,107 @@
...
+        try {
+            database.removeTask(todoListID, taskIndex);
+        } catch (DatabaseException e) {
+            e.printStackTrace();
+        }
+    }
+
+    /**
+     * Gets the task from the database
+     * DRAFT: I don't believe this method should exist really, I don't see how it would be used.
```
theo-damiani, line 91: I agree with you. I designed this method the previous week but I didn't really know what we will need. 
So yes, with the current development, I think we will not use it.

theo-damiani (approved):
Great works!

There is a lot of "controller things" in the current Item View.
I think we need to split and organize those. 

For example:
 - the todoAdapter (in ItemView) will be the view logic and just call the controller when he needs it. For now, it is autonomous and acts as a controller.
 - we also have methods like addTask which takes the input string and creates itself the new task. We need to maybe just perform a check on this string and call the right method of the controller.

So it is a great first implementation of the controller,
but one 'consequent' task for the next week will be to integrate the controller with the item view. And by doing it, we will better understand what we need in the controller I think.

VPellenc:
I removed the getTask, and added tests.

The coverage for this patch isn't at 100%, it should be at 82%. The reason being that e.printStackTrace() can't really be properly tested. I could add a test to run that code if needed but not sure if it's a good idea.
I'm open to other suggestions as well.

theo-damiani:
Hey, if you agree @VPellenc, I think we can close this PR.

VPellenc:
Yup

PR #47, opened 2021-03-18
view: create list selection layout

lgraziano:
This PR adds the list selection activity view. The buttons don't work yet but clicking on the list names opens the example list.


theo-damiani (approved):
Nice work!
I think it's enough for the Friday demo.

But I think it will be better to change the LinearLayout (which contains the todo lists) with a ListView as in the activity_item_view. 

So it will be easier to add to-do lists and integrate them into the code. 
We can also think to manage them with an adapter: it is much easier to recover which todo-list has been clicked and so launch the item view with the correct todo:
- The todo-list will be passed to the item view by putting it in an extra of the intent?
- You just passed the UUID of the todo list and recover it in the item_view from a cached database?


We need some tests for this PR, it's under the 80% threshold.

theo-damiani (approved):

shoutizix (approved):
Great job, I even tried it on my phone and it runs really well that's nice ! 
We'll need later to create a todoList with the "+" button at the bottom right of the screen and manage the lists but for now I think that's fine. Maybe we'll have to change the appearance a bit so that it shows great on every phone, maybe by adding lines between each list like if it was an excel's cell. That's just an idea.

PR #49, opened 2021-03-18
db: Create Database powered by Firebase Storage

YagoGG:
This acts as part II of the work started in #45, by adding an extra layer on top of the `FirebaseFileStorageService` that adapts its API to the `Database`.

It is important to note that the current calls are blocking (i.e. the main thread waits until the operation is fulfilled), which is clearly suboptimal but should be fine for now.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -44,6 +44,7 @@ dependencies {
     implementation 'com.google.firebase:firebase-storage'
     implementation 'com.google.code.gson:gson:2.8.6'
     testImplementation 'junit:junit:4.13.2'
+    testImplementation 'org.mockito:mockito-core:3.8.0'
     androidTestImplementation 'androidx.test.ext:junit:1.1.2'
     androidTestImplementation 'androidx.test.espresso:espresso-core:3.3.0'
     androidTestImplementation 'androidx.test:runner:1.3.0'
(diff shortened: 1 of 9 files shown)


shoutizix (approved):
Hey,

I think that's really great ! Indeed I will have to use it for #69 and it seems a solid basis on which I can work. I'm sorry but I just have a question about the TODO_LIST_PATH : Will this path be used by each users or it's the path of a big database ? 
For example if I connect with my account would it be stored on my account like : Noah/todo-lists/*
or on a big central database with all other users todo lists like : database/todo-lists/*

YagoGG:
Hey @shoutizix! Thanks a lot for reviewing and for the kind words. :)

The filesystem is indeed shared for all users, that's just how Firebase works. This is how it would look like for now:

```
├── /todo-lists/
│   ├── eeeeeeee-ffff-ffff-ffff-000000000000/
│        ├ eeeeeeee-ffff-ffff-ffff-000000000000.json
│        └ eeeeeeee-ffff-ffff-ffff-000000000001.json
│   ├── aaaaaaaa-bbbb-cccc-dddd-012345678901/
│        ├ ...
```
Where `eeeeeeee-ffff-ffff-ffff-000000000000` and `aaaaaaaa-bbbb-cccc-dddd-012345678901` would be user IDs, and `eeeeeeee-ffff-ffff-ffff-000000000000.json` the files containing the actual todo lists. Of course, the IDs are be random UUIDs.

Hope that helps, let me know if you have any other questions.
Cheers!
However, the access is already configured such that only each user is capable of accessing their own spaces.

sydneyhauke (approved):
Great work.

If I remember right, you explained that all operations from this new class are all synchronous. This will do for now and we could work on asynchronous later in a new sprint.

Issue #17, opened 2021-03-11
Make authentication persistent


YagoGG:
After using the Firebase login and establishing a proper session, creating the app will delete it (making it necessary to log back in).

PR #50, opened 2021-03-18
Propose login and logout to the user along with persistent authentication

sydneyhauke:
Addresses issues #30, #31, #28 and #17.

When the user is not logged in, the app shows the authentication activity. If the user presses the login button, the pre-built firebase UI pops up in order for the user to authenticate. Once the user is indeed authenticated, the list item activity is shown.  If the authentication somehow fails, an error message will show up and indicate to the user that authentication has failed. Inside the latter activity, a log out button is shown to the user in for her/him to log out. Pressing the button effectively logs out the user and shows the authentication activity again. Finally, if the user is logged in, stops/exists/restarts the app, its authentication is effectively persistent.

A few things to point out : 

- The logout button has been added to the list item activity without much thinking. I don't think it will stay there for a long time. to be refactored / placed someshere else later.
- The authentication activity is somewhat stark and very simple.
- The authentication persists but not infinitely over time. We should handle the case when an authentication token expires.


YagoGG (approved):
Hey @sydneyhauke! Super cool work, thanks a lot.

The code is pretty clean and I was able to test the feature locally. I only made a small nitpick/curious comment, but I'd be more than happy to merge.

theo-damiani (approved):
Looks great!

Tested and works fine, great jobs!

Issue #71, opened 2021-03-19
Create the custom adapter for the item view.


VPellenc:
Create an adapter like the one we use for the `Tasks` in the ÌtemViewActivity`.

Issue #69, opened 2021-03-19
Make TodoRepository asynchronous


YagoGG:
The current `FirebaseDatabase` implementation is blocking, which means that all database interactions stop the app's interface until the DB operation has finished.

We should add an intermediate layer that allows doing calling the operations synchronously, but that the actual work is done in the background (so the user can keep using the app while the DB is synchronized).

Issue #72, opened 2021-03-19
Convert the itemView to the MVVM schema


YagoGG:

YagoGG:
Closed in #76.

Issue #44, opened 2021-03-17
Keep a single source of truth for nested data


YagoGG:
Right now, objects in the device's memory are mutable and the modifications made to them (like adding a `Task` to a `TodoList`) will only be reflected in the database if we make the operation through the database (e.g. `Database.putTask`).

This is not good because it leads to having two sources of truth for the app's data (the one in memory, and the one in the database).

The original discussion about this started in #41, where you can find more information about this. More specifically, here's a more detailed explanation of the issue ([source](https://github.com/steroid-team/app/pull/41#issuecomment-801192257)):

> To give you an example of how can this go wrong:
>
> ```java
> Database database = new VolatileDatabase();
> TodoList todoList = new TodoList("My list");
> Task task = new Task("Do something");
> 
> database.putTodoList(todoList);
> todoList.addTask(task);
> 
> assertEquals(task, todoList.getTask(0));
> assertEquals(task, database.getTask(todoList.getId(), 0));  // This breaks
> ```
> 
> That is why I mentioned in [this comment](https://github.com/steroid-team/app/pull/36#issuecomment-799500539) that using an observer for each `TodoList` might be a good idea: that way, every time the in-memory object is mutated, we let the database know so they stay in sync. And, most importantly: that way we only have one way of doing an operation (instead of the two ones I show above).
> 
> So I totally agree with this having some important issues regarding mutability, but I think that we should figure out first what the view is going to consume, and then adapt the `Database`/`TodoList` APIs accordingly, because otherwise we are at risk of using a design that is not compatible with the Android view.
> 
> I would vote for merging this (functional yet quirky) initial version, create an issue to track what we are discussing, and give it a high priority in the next sprint so we can iterate over it. What do you think?

PR #76, opened 2021-03-19
model: MVVM pattern & LiveData

theo-damiani:
Hey everyone!

This is not really a pull request.
I just wanted to push it and create the pull request to explain some details.

I tried to define the MVVM model for the ItemViewActivity:
- for the model: nothing changes. It is the database, which I don't use for now in this branch (I will add it).
- for the view: it is still the ItemView but few things changed: the adapter just 'adapt' the to-do list, so it can be displayed by the ListView Layout. And all the modifications that a user can perform are taken by the view but performs by the view model.
- the view model is defined by the `ItemViewActivityViewModel` (maybe a too long name): it uses mutable live data. When the itemView activity is launched the model view will load the to-do list from the database (for now it is just a new to-do list). And performs any updates that the view sends to the model.

I am not really sure about all this stuff but from what the TA said this morning and my research after our meeting, it seems to be a good start with this new 'architecture'. And so Noah and Leandro get a first draft for their related task #71 and #69.

If you have any suggestions or anything, don't hesitate. 
I never used this stuff, so it is new to me. 

Cheers!


theo-damiani:
Make changes in the view:
- use a recycler view instead of a list view.
- create the todo adapter in a separate class.
- integrate the volatile database in the model but this leads to lots of problems as the database is not persistent.

The add task button no longer works. I will investigate this error.
But I think the pattern is well defined and you can get an idea to implement in the same way for the list selection view.

Also need to add tests.

theo-damiani:
Hey!

So I think the model is now good.
From the previous push, I fixed the Live Data updating. 
So it solved issue #44: I just need to take a look at the task class (adding a done field).
Also, the item view is fully adapted to the MVVM pattern #72.
Just to explain some details:
- Our current database API doesn't offer the possibility to return LiveData. In many MVVM explanations that I found, they always use a DAO (Data Access Object) and update the database (ROOM) with SQLite queries. So here, I just used the TodoRepository to update the database and take care of exposing only a LiveData instance even if it contains a MutableLiveData.
- This live data is observed in the Item View Activity. The observer takes care to update the TodoAdapter.
- The TodoAdapter just adapts a to-do list so it can be displayed in a recycler view (same as a List view apparently offer more functionality).
- This TodoAdapter notifies also when the data changed so the view is in a way 'rebuild'. The functionality works even if the adapter doesn't notify the view but the view will be less often updated.

I think we can maybe use this pattern for the list selection view, this concerns issues #71, #32.

Other things to note:
- the TodoRepository (maybe not the best name): should take care of getting the instance of the database in the case we have a persistent one. Here we just recreate each time a database.
- I added a public attribute `UUID id` in the Todo Repository to recover which todo list we are displaying in the item view as for now the list selection view doesn't give any info on which todo is selected.

I need to clean the code, remove the print (used to debug), modular some parts (CodeClimate complain of some too long function), and of course implement tests.

Cheers!

shoutizix (approved):
It's great and I definitely need it to start working on #69 ! I don't see any problem to my point of view but LiveData are also to new to me.

Cheers

YagoGG (changes requested):
Hey @theo-damiani!

Thanks a lot for the PR, it is really clever and the whole implementation is really cool.

It has taken me some time to fully understand the whole MVVM model, but now I feel comfortable with how we use it for the app.

I have left a few minor comments, but otherwise it looks great. Well done! :)

VPellenc (approved):
I read and reviewed the architectural changes, I didn't find any problem.
Other than what Yago already mentionned the patch looks good to me.

I didn't test the patch on my emulator that said, as I was away from it.

lgraziano (approved):
Great job ! I read the code and tested it, everything is just fine.

YagoGG (approved):
Looking great, thanks for the fixes and the amazing work overall @theo-damiani!

sydneyhauke (approved):
Nice work !

Issue #66, opened 2021-03-19
Add a rename function for the task lists


shoutizix:
We should be able to modify the name of task lists

This includes:
- Modifications in the view/activity.
- Corresponding backend.
- Tests.

Issue #59, opened 2021-03-19
Add feature to create task lists


YagoGG:
This includes:
- Creating the visual elements for adding tasks.
- Backend (plugged into the DB).
- Tests.

YagoGG:
Just FTR, in today's pre-sprint meeting, @lgraziano asked for a couple hours more for polishing. Make sure to add the proper "actual" label!

PR #77, opened 2021-03-21
Adding the rename function in the ListSelectionActivity

shoutizix:
Hey guys !

This addresses the issue #66 I had to refactor a bit the code and implement an adapter like we used for the ItemViewActivity class. I have also changed the ScrollView that contained TextView by a ListView containing TextView since it was better to get which view was cliked on and modify/open the corresponding TodoList. 
I think that #59 should be pretty easy using the adapter I've added in this PR. 

I still have to add tests and use the Database but otherwise the rename functionality should work great !


shoutizix:
Now all should be great !
I'm waiting for your reviews :)

app/src/main/java/com/github/steroidteam/todolist/ListSelectionActivity.java
```diff
@@ -1,27 +1,130 @@
...
+
+        @Override
+        public Object getItem(int i) {
+            return todoLists.get(i);
+        }
+
+        @Override
+        public long getItemId(int i) {
+            return i;
+        }
```
YagoGG, line 81: Shouldn't this ID be the list's ID? Something like `todoLists.get(i).getId()`.

app/src/main/res/layout/layout_todo_list_item.xml
```diff
@@ -0,0 +1,16 @@
...
+
+    <TextView
+        android:id="@+id/layout_todo_list_text"
+        android:layout_width="0dp"
+        android:layout_height="55dp"
+        android:layout_marginTop="7sp"
+        android:layout_marginEnd="7sp"
+        android:layout_weight="1"
+        android:fontFamily="sans-serif-black"
+        android:text="TextView"
```
YagoGG, line 14: Not super sure about this, but I believe that if `"TextView"` is just a placeholder, we can just remove the whole `android:text` attribute. That way we avoid useless string (that are also considered hardcoded by the linter).

YagoGG (changes requested):
Hey @shoutizix! Thanks a lot for the PR.

I left a couple inline comments about the code.

As for the feature, it looks super neat! I have been able to try it out in my emulator and it worked like a charm. I found a small issue, though:

Whenever you try to rename a list, an alert pops up asking you for the new name. However, the input field allows multi line inputs, which leads to weird behaviour:
    image
    image
    image

shoutizix:
Hey !

You're absolutely right on the points that you've mentioned, I've made the corresponding changes.

app/src/main/java/com/github/steroidteam/todolist/ListSelectionActivity.java
```diff
@@ -1,27 +1,131 @@
...
+            return true;
+        });
+    }
+
     public void openList(View view){
         Intent itemViewActivity = new Intent(ListSelectionActivity.this, ItemViewActivity.class);
         startActivity(itemViewActivity);
     }
+
+    private class todoListAdapter extends BaseAdapter{
```
theo-damiani, line 60: Maybe it is better to use a recycler view and a holder. (better performance and more functionality)
And we can maybe put it in another file so it can be re-used if necessary and it's better for understanding the global structure of the app.

theo-damiani (approved):
Great work!

I just mentioned two little things, but for the first one (regarding the UUID instead of the title), I think we will address it when we will implement the view model for this activity, for now, without it, you can't really do differently

app/src/main/java/com/github/steroidteam/todolist/ListSelectionActivity.java
```diff
@@ -1,27 +1,131 @@
...
+        }
+
+        @Override
+        public Object getItem(int i) {
+            return todoLists.get(i);
+        }
+
+        @Override
+        public long getItemId(int i) {
+            return todoLists.get(i).getId().getLeastSignificantBits();
```
YagoGG, line 80: Why are we using this instead of the ID? :thinking:

YagoGG (commented):
Hey @shoutizix! Thanks a lot for the changes.

I think this is about to be ready to merge. I just have a question regarding the significant bits trick, and I also share with @theo-damiani the preference for `UUID`s instead of titles (which, unlike `UUID`s, might not be unique).

Cheers!

YagoGG:
Please do not forget to update the branch with the latest `master` to make sure that there are no merge conflicts. Thanks a lot! :)

app/src/main/java/com/github/steroidteam/todolist/ListSelectionActivity.java
shoutizix: Because the ID is an int and I've made some research and it seemed a used way to do but we could certainly change that in the future

shoutizix:
Yes sure sorry I resolve the conflict now it should be good to merge !

lgraziano (approved):
I read the code and tried it, it works perfectly ! Very nice job 👌

app/src/main/java/com/github/steroidteam/todolist/ListSelectionActivity.java
YagoGG: So just to be clear, is this to cast the ID from `int` to `long`? Because if that's the case, you can just do:

```java
(long) todoLists.get(i).getId()
```

shoutizix: well yes it would make sense and it's more understandable I'll do it rn

YagoGG: Makes sense, thanks for the clarification!

theo-damiani:
Again maybe the alert dialog will cause some errors. But works fine for me.

app/src/main/java/com/github/steroidteam/todolist/ListSelectionActivity.java
```diff
@@ -1,43 +1,154 @@
...
+        todoLists.add(new TodoList("Some stuff"));
+        setTitle("TODO Lists");
+
+        ListView listView = findViewById(R.id.activity_list_selection_itemlist);
+        adapter = new todoListAdapter(todoLists);
+        setListViewSettings(listView);
+        //Toolbar toolbar = findViewById(R.id.list_selection_toolbar);
+        //setSupportActionBar(toolbar);
+        //ActionBar actionBar = getSupportActionBar();
+        //actionBar.setDisplayHomeAsUpEnabled(true);
```
YagoGG, line 55: Looks like we're not using this anymore, maybe we can remove it?

YagoGG (approved):
Hey, @theo-damiani!

Thanks a lot for the changes; I just did a second batch of functional tests and the new feature now works like a charm.

I left a minor comment, and the linter is also complaining about a couple hardcoded strings & missing attributes. Not super important, but it'd be great if we could take care of those before merging. LGTM otherwise! :+1: 

Cheers!

PR #78, opened 2021-03-22
Feature/list selection controller

lgraziano:
The controller for the list selection activity, I don't know if it is still relevant because of the switch to MVVM. If not let me know and I'll delete it. Also I couldn't get 100% of code coverage because of the try catch blocks. If you have a solution it is welcome !


YagoGG:
Hey @lgraziano! Thanks a lot for the PR.

I would definitely be more comfortable if @theo-damiani could confirm this, but for what I have understood after checking out the code you are right, this would be superseded by #76.

theo-damiani (commented):
Hey! 

The code will be useful for the the view model (it's kind of a controller). We can maybe keep the code, it will be easier to implement the view model but maybe not merged this PR in the master. Maybe leave the PR open and we will close it when the view model will be ready.

PR #80, opened 2021-03-24
Implent and test user concept with firebase as backend

sydneyhauke:
Addresses #67 on user concept in the app.
To be honest I took inspiration with Firebase's authentication system in order to propose some useful features we might need later in the project. User profile fetching, email update, password reset etc.

The idea here is to abstract away the concept of User. In the end, the FirebaseUser object from Firebase Auth is wrapped around so that we don't spread dependencies all over the codebase to Firebase Auth.


shoutizix (commented):
Hey, 

It seems a great feature to have nice done ! 

I just don't really understand the test noExceptionWhenLoggIn(), because at first you use "when" with a firebaseUserMock but right after in the try catch segments you use fBaseUser and test the same method as before, why ?

app/src/main/java/com/github/steroidteam/todolist/user/User.java
```diff
@@ -0,0 +1,47 @@
...
+            throws UserLoginException;
+    void sendEmailVerification(OnCompleteListener<Void> onComplete)
+            throws UserLoginException;
+    void setUserPassword(String newPassword, OnCompleteListener<Void> onComplete)
+            throws UserLoginException;
+    void sendPasswordResetEmail(OnCompleteListener<Void> onComplete)
+            throws UserLoginException;
+    void deleteUser(OnCompleteListener<Void> onComplete)
+            throws UserLoginException;
+    void reauthenticate(OnCompleteListener<Void> onComplete);
```
theo-damiani, line 30: Maybe this method could also throw a `UserLoginException`

theo-damiani (approved):
Great job!

Just one quick remark, I think your `stateListener` acts as an observer. So do you think it will be better to rename with `stateObserver` and the same for the concerning methods? So we can see that you indeed use this pattern.

theo-damiani:
Hey @sydneyhauke!

I just ran and applied the spotless check.
So just I more review and we can merge your PR.

app/src/main/java/com/github/steroidteam/todolist/user/User.java
YagoGG: Why would we want such a method?

I guess that if the user's token has expired they will have to log back in to get a fresh one, and that is basically running the whole authentication flow as if they were logged out.

app/src/test/java/com/github/steroidteam/todolist/user/FBaseUserTest.java
```diff
@@ -0,0 +1,164 @@
...
+                });
+    }
+
+    @Test
+    public void noExceptionWhenLoggIn() {
+        FirebaseUser firebaseUserMock = Mockito.mock(FirebaseUser.class);
+        when(authMock.getCurrentUser()).thenReturn(firebaseUserMock);
+
+        Task<Void> taskMock = Mockito.mock(Task.class);
+        // when(firebaseUserMock.updateProfile(any())).thenReturn(taskMock);
```
YagoGG, line 100: Do we want this?

app/src/test/java/com/github/steroidteam/todolist/user/FBaseUserTest.java
```diff
@@ -0,0 +1,164 @@
...
+        when(firebaseUserMock.updateEmail(anyString())).thenReturn(taskMock);
+        when(firebaseUserMock.sendEmailVerification()).thenReturn(taskMock);
+        when(firebaseUserMock.updatePassword(anyString())).thenReturn(taskMock);
+        when(firebaseUserMock.getEmail()).thenReturn("");
+        when(authMock.sendPasswordResetEmail(anyString())).thenReturn(taskMock);
+        when(firebaseUserMock.delete()).thenReturn(taskMock);
+
+        OnCompleteListener<Void> onComplete =
```
YagoGG, line 108: This can be replace by a lambda.

app/src/main/java/com/github/steroidteam/todolist/user/User.java
```diff
@@ -0,0 +1,54 @@
...
+    void sendEmailVerification(OnCompleteListener<Void> onComplete) throws UserLoginException;
+
+    void setUserPassword(String newPassword, OnCompleteListener<Void> onComplete)
+            throws UserLoginException;
+
+    void sendPasswordResetEmail(OnCompleteListener<Void> onComplete) throws UserLoginException;
+
+    void deleteUser(OnCompleteListener<Void> onComplete) throws UserLoginException;
+
+    void reauthenticate(OnCompleteListener<Void> onComplete);
```
YagoGG, line 35: I am not a big fan of callbacks; personally I think that returning `CompletableFuture`s is better here.

In any case it is functional and not a major issue, just a matter of being as consistent as possible with the asynchronous code. So I would be okay with leaving this as-is if it is a big hassle.

app/src/main/java/com/github/steroidteam/todolist/user/User.java
```diff
@@ -0,0 +1,54 @@
+package com.github.steroidteam.todolist.user;
+
+import com.google.android.gms.tasks.OnCompleteListener;
+
+public interface User {
+    /* Trivial method to have here */
+    boolean isLoggedIn();
```
YagoGG, line 7: I do not understand under which circumstances could `user.isLoggedIn()` be `false`. If the user has not authenticated, should not `user` be `null`?

That is what I would expect, especially considering that actual authentication takes place upon instantiation of the class (if I understood correctly).

YagoGG (changes requested):
Hey @sydneyhauke!

Thanks a lot for the PR, looks really good.

I left a few comments, mostly regarding the `User `API. Since they are quite subjective, I would love to know your thoughts on them as well.

Cheers!

YagoGG:
By the way, have you considered following the singleton pattern for the `User` class?

Considering that it will be unique within the session, it might be useful to do it that way. I am currently working with a similar POC for one of my pull requests, and I think that it could fit great here.

app/src/main/java/com/github/steroidteam/todolist/user/User.java
sydneyhauke: Indeed, according to Firebase's documentation the returned user object should be non-null once a user is authenticated. However the authenticated user has a limited lifetime (expired authentication token, logged out etc)

app/src/main/java/com/github/steroidteam/todolist/user/User.java
sydneyhauke: You're right. I added these callbacks by following the way the Firebase authentication's API works.

YagoGG:
Hey @sydneyhauke! I believe a couple of comments have not been addressed yet.

Please feel free to request a new review when it's ready to go, I'll be glad to give it a go! :)

lgraziano (approved):
Great work, I think we can merge it right away 👍

YagoGG (approved):
Thanks for the changes, @sydneyhauke! LGTM.

I created #171 as a followup issue to fully integrate this into our authentication flow, just in case you might be interested.

Cheers!

PR #83, opened 2021-03-24
Feature/notes database

VPellenc:
This also has the content of my other PR (the view) as it is a required dependency.

This implemented the Note related actions in the VolatileDatabase, and added a minimal (and to be replaced later) integration with the view.

What you can do in this draft:
- See a list of notes
- Click on one note to go to the NoteDisplayActivity to see its content
- Modify the content of a note, click on the save button, and observe it has indeed been modified in a VolatileDatabase

What am I going to do:
- First add the tests for what is already done
- If I have time, implement the ability to create a Note, and to change the title of an existing Note


theo-damiani (approved):
Great work!

Tested locally and looks good!

Just one thing to mention:
You use a listView layout and an implementation of the BaseAdapter which is great and we did this last time for the item view activity but after my research and work on the MVVM pattern, I saw that the recycler view was more a common implementation to do. 
Apparently, it's a modernized version of the listView and gridView, it's getting better performance, and offers more functionality. I used very simply the recycler view when converting the item view to MVVM. 
Maybe it can be a task for the next sprint to add a recycler view instead of a list view.
And also create the ViewModel for the notes view. To use MVVM.

VPellenc:
Yes, that makes sense, let's do that in the next sprint.
For this PR I will probably just add the tests and then focus my time to do some reviewing.

VPellenc:
There should be one more test, to check that clicking on the save button actually makes the Note persist, however I don't have the time to add it before 7pm so I ready the PR for review now.

sydneyhauke (changes requested):
Ciao,

Works great !

Before merging, make sure the linter issues are fixed. In particular those with hardcoded strings.

YagoGG:
@VPellenc, could you please solve the merge conflicts?

Pinging @sydneyhauke, as I believe that the changes requested have already been taken care of.

VPellenc:
I believe the work of this PR is going to be more or less replaced by #151 so it looks to me we should probably close this PR.
I don't know how that will stand with respect to #149 however since originally my PR was considered as a dependency. Do we now consider #151 to be the dependency to #149 ?

theo-damiani:
> 
> 
> I believe the work of this PR is going to be more or less replaced by #151 so it looks to me we should probably close this PR.
> I don't know how that will stand with respect to #149 however since originally my PR was considered as a dependency. Do we now consider #151 to be the dependency to #149 ?

I need that the database handle note so I think yes.

sydneyhauke:
Let's close this PR. issue #151 already handles that.

PR #86, opened 2021-03-25
Feature/async todorepository

shoutizix:
Hey guys !

For now I was only able to make the putTask and the removeTask asynchronous, but on my smartphone it works great. We definitely will have to make asynchronous ALL the database part in the other sprints, but for this one I think this is already something that we could merge. 
Let me know your opinions ! :)


shoutizix:
Codeclimate is detecting code duplicate but in fact it's not the same as it is the attributes of 2 differents classes

sydneyhauke (commented):
Ciao,

I think we might be going the wrong way in dealing with asynchrony.
Currently we are wrapping asynchrony around a synchronous database class. Here all (synchronous) calls to the database are put into asynchronous tasks.

I think we should do otherwise because creating a concrete AsyncTask for each operation seems cumbersome. Each time we add a new operation we will have to declare a new concrete task.

Instead, I suggest we follow the way Firebase handles asynchrony, with futures. In the Database class and its concrete classes, return futures and don't block on them ! In order to handle the result of a future, call the .thenApply() or .andThen() methods on the returned future out of the database.

Let me know if you want to further discuss this.

shoutizix:
Hey @sydneyhauke thanks a lot for the review !

This PR is linked to the issue #69 and I think that the description was to exactly to be able to call the operations synchronously but there will be done in background, in this way I think that what I've done is right. 

But it's true that to create an AsyncTask every time isn't the most optimal way to do that but it was the easier way to begin with.

I also think that we should consider for the next sprints what you've mentionned to use futures.

YagoGG:
As per the team standup (12/04/2021), this is being superseded by @sydneyhauke's work on data persistence.

PR #111, opened 2021-03-26
Feature/item modification

lgraziano:
When long pressing on a task, a dialog box allows to choose between modifying or deleting the task. We can make these actions more intuitive in the future.


theo-damiani:
I'm closing this PR as we merged #112

PR #112, opened 2021-03-27
Feature/Task Renaming & Deletion improvement (with the work of Leandro: PR111)

theo-damiani:
Hey!

Thank you to Leandro for his work. 
(so maybe we can close his PR #111, his work is in this one also)
I just adapt it and tried to make it as close as possible to what Yago showed us last time: 

You have to long click on a task and it opens an edit text. So you can rename it if you want. You close the modification layout by pressing one of the two buttons. The green one is to save your modification and the red is to delete the task.

Just one remark to think of. It's that whenever you push the green button you go through the view model, the repository, and the database, even if you didn't change anything so maybe I will add a check to verify the user indeed rename the task. 

And also I removed the deletion from this button, I don't know if you prefer to keep it also:

### DIFF
--- a/app/src/androidTest/java/com/github/steroidteam/todolist/ItemViewActivityTest.java
+++ b/app/src/androidTest/java/com/github/steroidteam/todolist/ItemViewActivityTest.java
@@ -179,7 +179,7 @@ public class ItemViewActivityTest {
         return new BoundedMatcher<View, RecyclerView>(RecyclerView.class) {
             @Override
             public void describeTo(Description description) {
-                description.appendText("View holder at position " + String.valueOf(position) + " ");
+                description.appendText("View holder at position " + String.valueOf(position) + ", expected: " + expectedText + " ");
             }
 
             @Override
--- a/app/src/androidTest/java/com/github/steroidteam/todolist/ItemViewActivityTest.java
+++ b/app/src/androidTest/java/com/github/steroidteam/todolist/ItemViewActivityTest.java
@@ -147,42 +147,8 @@ public class ItemViewActivityTest {
                     RecyclerViewActions.actionOnItemAtPosition(
                             0, MyViewAction.clickChildViewWithId(R.id.layout_task_delete_button)));
 
-            onView(withText("You are about to delete a task!")).inRoot(isDialog()).check(matches(isDisplayed()));
-
-            onView(withText("Yes")).inRoot(isDialog()).perform(click());
-
             //after deleting the first item we check that we have the second one at position 0.
             onView(withId(R.id.activity_itemview_itemlist)).check(matches(atPositionCheckText(0, TASK_DESCRIPTION_2)));
-        }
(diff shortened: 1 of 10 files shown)


YagoGG (changes requested):
Hey @theo-damiani! Thanks a lot for the PR, and props as well to @lgraziano for the great work :tada:.

I found a _double issue_ while testing the feature. Here are the steps to reproduce:
1. Start the app & log in using any auth provider.
2. Enter the "Daily todo" list.
3. Create two tasks with any text, like this:
    example
4. Long-press the second task, so it enters in "edition mode".
5. Long-press the first task, so it also enters in "edition mode". It should be looking like this:
    example
6. Hit the red button next to the first task to delete it.
7. As you will notice, the second task **is no longer in "edition mode"**.
8. Create a new task.
9. As you will notice, the new task **is unexpectedly in "edition mode"**.

In any case, I feel like it might be more intuitive if we can only edit one task at the time. That might make this problem easier to solve.

Also, in my personal opinion, here are 2 tings I miss:
- I would prefer if I could start "edit mode" just by tapping in the task (instead of long pressing).
- It would look better if we could keep the same font size in the task description when it enters "edit mode". We could remove the line that appears under the text field when it is editable, so we don't have space issues.

Of course this is quite opinionated, so I would love to hear your thoughts as well about this. :)

Cheers!

YagoGG:
> And also I removed the deletion from this button, I don't know if you prefer to keep it

+1 to removing it.

theo-damiani:
Thank you for your review Yago!

Sure you right it's better to be able to modify only one task at a time.

And yes I wanted to have the same font size but I didn't know it was possible to remove the underline!

I will push the changes!

YagoGG (approved):
Hey @theo-damiani! Awesome, thanks a lot.

I also took the liberty to push a couple extra commits, I hope you don't mind. I believe 079de72 could improve the user experience when modifying tasks, and it was also a great exercise to get familiarized with the new code :)

Let's get this baby merged! :rocket:

theo-damiani:
Thanks! Yes, I tested your commit 079de723380c248de8a1e7443d48ce4964f051b7 and it really improved the user experience, good feature !

lgraziano (approved):
Very nice work ! I just have two small remarks :
- since the two buttons are next to each other, I am worried one could delete a task when trying to confirm a change. Maybe that's not really an issue, let me know what you guys think about it.
- There's a bug with the ticks, they sometime "jump" to the next task when modifying a task. Still, I think we'd better merge this PR now and fix that later introducing "ticked" as an attribute of a task like we discussed.

Except for that everything works fine and the tests pass correctly 😃👍

YagoGG:
+1 to get this merged for now.

@lgraziano, could you please file an issue with the steps to reproduce the behaviour with the checkboxes? That way we can keep track of it in the future.

Cheers!

PR #113, opened 2021-03-30
tests: Add CI check for Google Java code style

YagoGG:
This makes sure that out code looks pretty and consistent across files.

It includes a new Gradle task that automatically reformats the code, so we can write without worrying about the format. The only thing we need to do is to run the `verification/spotlessApply` check before commiting new code!

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -39,6 +39,8 @@ check_android_task:
     ./cc-test-reporter upload-coverage
   lint_script:
     ./gradlew lintDebug
+  code_format_script:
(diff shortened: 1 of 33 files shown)


YagoGG:
Just as a heads-up, it looks like it's a huge PR but the changes in the Java files are just automatic reformatting to comply with the new rules. They do not include any functional changes!

theo-damiani (approved):
Thanks for these changes!
Definitively a good feature to help us having a clean code!

YagoGG:
Updated with the latest `master` and merged. Thanks, folks!

PR #117, opened 2021-04-01
Refactor/package structure

theo-damiani:
Refactor the packages to have structure well defined:


theo-damiani:
I just don't know if we should merge this now or wait for example Friday that all our work is merged and after perform the refactoring. Otherwise, there will be too much refactor for each of us.

sydneyhauke (approved):
Very nice indeed. It will tidy up a bit the project.

lgraziano (approved):
That's much cleaner, nice work !
I guess we can marge that right now

PR #119, opened 2021-04-01
ListSelectionView MVVM

theo-damiani:
Hey!
So in this PR I switch the list selection view to MVVM.

I ran into one small problem: as in the VolatileDatabase we use a hash map, when you insert a new todo, it is not displayed in the right order, I think it is the order of the UUIDs. So I just need to think of how re-order the element. This just adds a little more work.

And also I think, I will try to implement renaming and deletion as we did for the tasks.


YagoGG:
Hey @theo-damiani!

I'll try to have a look at the PR as soon as possible, but I saw this and I thought it could be worth discussing this:

> [...] when you insert a new todo, it is not displayed in the right order [...]

In my opinion that's not a big problem. Keep in mind that in the future, once todo lists and notes have the proper attributes, we'll be able to sort them by more consistent criteria (in fact, we can even let the user choose the sorting method).

I would not waste a lot of time trying to keep them in creation order because that will be something very relative once we have synchronization properly set up.

As an example to illustrate what I mean, imagine that a user creates two lists in different devices. What's the order in which they should appear in each device? Whether it's alphabetical, chronological, or any other kind, it's something that should be based in the item's attributes, and not how they were populated into the database.

Just my 2 cents, hope that helps!

theo-damiani:
Hey!
If someone understands why the tests are failing, feel free to add a comment or change by yourself the code.
I really don't understand this issue.
Thanks!

theo-damiani:
Without the failing tests we go down to ` codeclimate/total-coverage — 74% (-10.5% change)`

YagoGG:
Hey @theo-damiani!

I have been taking a look at the CI issue, but after a couple hours I had no luck so far. :(

I have added a commit that dumps the `logcat` to the CI's artifacts, so we can inspect the CI runs more easily.

I will try to have another look at it over the next few days, but hopefully that will help you debug in the meantime.

Cheers!

theo-damiani:
Thanks, @YagoGG!
I think I found the error. The tests were failing because the dialog when adding/renaming a todo was out of the screen or maybe just the edit text in the dialog.
It seems to work now, I will add the other tests tomorrow, and I will clean up the code and maybe add some design to the dialogs.

VPellenc (approved):
Quite a large and well-tested PR.

Right now entering the ListSelectionActivity creates a new ViewModel which creates a new TodoRepository which creates a new VolatileDatabase it uses to manage the Todolists. This means that if we return to the ListSelectionActivity after going to the Notes for example then we lose the changes. Of course in itself that is not a problem as we still don't have the persistency with the Firebase database but it still makes me wonder about how we should use the VolatileDatabase.
Do we want to create a new instance everytime we go in a specific location in the app and then load again the data from Firebase, or having some sort of singleton implementation created upon app launch and staying until the app is closed could make sense?
For example in my Feature/notes database PR, although it was a bit quick and dirty as it would need to be changed again in the future anyway, the VolatileDatabase is statically available from every part of the program, so some sort of persistency is possible that way.

VPellenc:
> > Quite a large and well-tested PR.
> > Right now entering the ListSelectionActivity creates a new ViewModel which creates a new TodoRepository which creates a new VolatileDatabase it uses to manage the Todolists. This means that if we return to the ListSelectionActivity after going to the Notes for example then we lose the changes. Of course in itself that is not a problem as we still don't have the persistency with the Firebase database but it still makes me wonder about how we should use the VolatileDatabase.
> > Do we want to create a new instance everytime we go in a specific location in the app and then load again the data from Firebase, or having some sort of singleton implementation created upon app launch and staying until the app is closed could make sense?
> > For example in my Feature/notes database PR, although it was a bit quick and dirty as it would need to be changed again in the future anyway, the VolatileDatabase is statically available from every part of the program, so some sort of persistency is possible that way.
> 
> For me when you enter the ListSelectionActivity, it creates a ViewModel. The ViewModel creates a TodoRepo but the TodoRepo should only instantiate a new volatile database if there isn't already one (singleton pattern). But as we don't have persistency yet, I didn't implement this. For me, the TodoRepo takes care to manage the data between Firebase and the Room/Cached database.
> So the ViewModel doesn't need to know from where the data is loaded.
> So I was thinking that when you enter the app or an activity for the first time the TodoRepo load data from firebase but after the data is stored locally. (We don't perform calls to firebase every time) But we need to talk about the async and room aspect of the database. for sure!

Fair enough.

app/src/main/res/layout/alert_dialog_input.xml
```diff
@@ -0,0 +1,18 @@
...
+
+    <EditText
+        android:id="@+id/alert_dialog_edit_text"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:inputType="text"
+        android:maxLines="1"
+        android:backgroundTint="@color/dark_grey"
+        android:textCursorDrawable="@drawable/custom_cursor"
+        android:hint="..." />
```
YagoGG, line 16: Maybe we can replace this with something more descriptive, like `My todo list` or `Your new description`?

Also, it should be a `@string` resource as the linter points out.

YagoGG (changes requested):
Hey @theo-damiani! Thanks a lot for the great work.

I noticed that you are using `CardView`s in the layout for todo lists, which is different to the custom layout that we use in [`layout_note_item.xml`](https://github.com/steroid-team/app/blob/master/app/src/main/res/layout/layout_note_item.xml), which has some implications:

1. The box drops shadows around them, which look kind of awkward with out current style (especially with the round corners):
todo list item
2. The name of the todo list is no longer vertically centered. This is how it used to look:
old todo list item
3. Most importantly, our components in the todo/note selection views are no longer consistent. I am okay with using `CardView`s if we consider that's best, but in that case we should be using it both in the todo selection view AND the note selection view. It could also be a good idea to create a custom component ourselves, that is reused/extended in both views (like our own version of a `CardView`, basically).

I really liked the concept of sliding left/right for the actions, but I was a bit confused at the beginning about what did each direction do. I don't know how hard would it be to show a small hint when swiping; the Gmail app does something like this, but maybe something like a small text or just a colour (red for deleting, orange for renaming) could do the trick as well:

sliding in gmail

Again, I have no idea whatsoever on how hard the implementation would be, so I am okay with merging this feature as-is and adding the UX enhancements in the backlog.

Thanks again for the great work!

theo-damiani:
> 
> 
> Hey @theo-damiani! Thanks a lot for the great work.
> 
> I noticed that you are using `CardView`s in the layout for todo lists, which is different to the custom layout that we use in [`layout_note_item.xml`](https://github.com/steroid-team/app/blob/master/app/src/main/res/layout/layout_note_item.xml), which has some implications:
> 
>     1. The box drops shadows around them, which look kind of awkward with out current style (especially with the round corners):
>        todo list item
> 
>     2. The name of the todo list is no longer vertically centered. This is how it used to look:
>        old todo list item
> 
>     3. Most importantly, our components in the todo/note selection views are no longer consistent. I am okay with using `CardView`s if we consider that's best, but in that case we should be using it both in the todo selection view AND the note selection view. It could also be a good idea to create a custom component ourselves, that is reused/extended in both views (like our own version of a `CardView`, basically).
> 
> 
> I really liked the concept of sliding left/right for the actions, but I was a bit confused at the beginning about what did each direction do. I don't know how hard would it be to show a small hint when swiping; the Gmail app does something like this, but maybe something like a small text or just a colour (red for deleting, orange for renaming) could do the trick as well:
> 
> sliding in gmail
> 
> Again, I have no idea whatsoever on how hard the implementation would be, so I am okay with merging this feature as-is and adding the UX enhancements in the backlog.
> 
> Also, please have a look at the linter warnings for the new code.
> 
> Thanks again for the great work!

Thanks for the review!
I made some changes. But for the alignment of the title in a view, it is the same as for the note:



And I didn't know how to center it vertically has in fact you have the recycler view to print some tasks:

theo-damiani:
The icons are available under the Apache Licence Version 2.0: Copyright [2021] [name of copyright owner]
I found them in vector assets of Android Studio.

Small picture to understand the bounds for drawable:

theo-damiani:
reminder: remove logcat from CI artifacts

app/src/main/java/com/github/steroidteam/todolist/view/ListSelectionFragment.java
```diff
@@ -1,151 +1,156 @@
...
+                    titleInput.getText().clear();
+                    dialog.dismiss();
+                });
+        builder.setNegativeButton(
+                getString(R.string.cancel),
+                (DialogInterface dialog, int which) -> {
+                    dialog.dismiss();
+                });
+        Dialog dialog = builder.show();
+        dialog.getWindow().setGravity(0x00000035);
```
YagoGG, line 100: What's this hex number? If we cannot replace it with something else (e.g. a `Gravity.*` constant or a value in `dimens.xml`), it'd be great to have a comment explaining it.

YagoGG (changes requested):
Hey @theo-damiani! Thanks a bunch for the changes.

Congrats on the dialog's styling, it looks amazing!

I think I misunderstood the changes in the list selection itself, hence my confusion when the title was misaligned (my lists were empty). I see now that you decided to display the tasks both in the `ItemViewFragment` as well as in the `ListSelectionFragment`.

Personally I do not think it's super intuitive from the UX point of view; I would keep it as before (i.e. show only the lists' titles), but of course it's a subjective matter.

In any case, I ran into the following issue when trying to rename/delete lists: the dialog appears under the dimmed background.

This is how the app looks upon startup: 
list selection

And this is what I get if I slide right to rename:
list renaming dialog

I would expect everything to be dimmed, *except* the dialog itself (which should remain white).

I noticed this also happens in #209, so I guess that it's related to how we create the `Dialog`. You might want to consider a [`DialogFragment`](https://stackoverflow.com/questions/13765127/dialogfragment-advantages-over-alertdialog/13765411), though I am not sure if that would solve the problem at hand.

Please let me know if you need any further details or help! :)

YagoGG (approved):
Looks brilliant, great work @theo-damiani!

I have taken the liberty to push 2 trivial commits, I hope that's okay. Cheers!

PR #121, opened 2021-04-01
Show Current GPS Location + search Location

shoutizix:
I had troubles to make it works with maps API, so for now I've only added a new button in the toolbar of the notes "Add location" and it only displays the map centered at Sidney. 
I have to also hide/encrypt the google API key but I don't know how to do it so that CI would still be able to run test that needs it.
I have to show the position of the user at first and maybe add a feature to be able to search a place and it will point to it.


shoutizix:
Hey,

So first to access the MapsActivity we just have to access the note part and then click on a note and click on the note_header.

It will ask for permission to access the location of the user and if the user doesn't give permission or have the localisation disabled it will put a marker at a default place (right now it's Sydney) and center the camera to it.

If the user gave permission and have localisation enabled, then it will put a marker at the current place of the user.

The tests work fine but the coverage is very low, it only includes onCreate method, and I think it's because all the other methods are called in onMapReady but onMapReady isn't called form the activity, it's only the callback of getMapAsync method at the end of onCreate.
Because of that I don't know how to increase the code coverage, so if you have any idea please let me know !

app/src/main/res/layout/activity_maps.xml
```diff
@@ -0,0 +1,26 @@
...
+
+    <SearchView
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:id="@+id/sv_location"
+        android:queryHint="Search..."
+        android:iconifiedByDefault="false"
+        android:layout_margin="10dp"
+        android:elevation="5dp"
+        android:background="@drawable/rounded_corner_bg"
```
theo-damiani, line 24: The layout hasn't the rounded corner:

app/src/main/java/com/github/steroidteam/todolist/view/MapsActivity.java
```diff
@@ -0,0 +1,261 @@
...
+                location.substring(0, 1).toUpperCase() + location.substring(1);
+        if (marker != null) {
+            marker.setPosition(latLng);
+            marker.setTitle(locationWithUppercase);
+        } else {
+            marker =
+                    map.addMarker(
+                            new MarkerOptions().position(latLng).title(locationWithUppercase));
+        }
+        map.animateCamera(CameraUpdateFactory.newLatLngZoom(latLng, 10));
```
theo-damiani, line 120: It will be great to have a private final attribute for the int 10, as you did for `DEFAULT_ZOOM `

app/src/main/java/com/github/steroidteam/todolist/view/MapsActivity.java
```diff
@@ -0,0 +1,261 @@
...
+        }
+    }
+
+    private void deviceLocationMarkerLocNotNull(Location location) {
+        map.moveCamera(
+                CameraUpdateFactory.newLatLngZoom(
+                        new LatLng(location.getLatitude(), location.getLongitude()), DEFAULT_ZOOM));
+        if (marker != null) {
+            marker.setPosition(new LatLng(location.getLatitude(), location.getLongitude()));
+            marker.setTitle("I'm here");
```
theo-damiani, line 183: Also for a string like this maybe you should extract it and add a resource in strings.xml.
Then you can recover it in the activity with `getString(R.string.your_string_name);`

theo-damiani (approved):
Hey @shoutizix,
Great job!

I tested your PR and it works really well!
I just have some remarks:

- Main problem: when you enter a fake location, for example just one character, this makes crash the application: 


Maybe you just have to add a check in the method `onQueryTextSubmit`

- I wanted to say that it should be great to have a button to go in the map activity and to be able to display (or just store) the location in a note, but I saw that this is your task for sprint 6. So for me, it is good to merge it like this.


If you can just fix the issue when you enter a fake location, it would be perfect!

app/src/main/java/com/github/steroidteam/todolist/view/MapsActivity.java
shoutizix: That's true I forgot to do it !

app/src/main/java/com/github/steroidteam/todolist/view/MapsActivity.java
shoutizix: Yes you're completely right ! Thanks a lot for pointing this out I've corrected it right now

shoutizix:
Hey @theo-damiani thanks a lot for the review !

You were absolutely right on the points that you've mentioned and I've made the corresponding changes.

theo-damiani:
> 
> 
> Hey @theo-damiani thanks a lot for the review !
> 
> You were absolutely right on the points that you've mentioned and I've made the corresponding changes.

Awesome! 
For me it is ready to be merged !

lgraziano (approved):
The code looks fine and the map works really well, nice work !
I just had some problem when denying the permission to access the localisation, the map doesn't open and it takes me back to the notes screen instead. However, since the main feature works I suggest we merge that now and fix the issue in the next PR

PR #145, opened 2021-04-06
Feature/check task

theo-damiani:
Hey!
So in this PR, I added the logic to manage the boolean Done of a task.
I also refactor the way we delete/modify tasks because, in a previous way, there are a lot of problems when you scroll for example if you have a lot of tasks.
And also now when you click on a task, it opens a small panel to update it. And so it will be easier to add features for the future. For example, I wanted to add a boolean "prioritize" so when a task has this flag it is displayed on the list selection view.

I also try to add design according to what Yago did.

### DIFF
--- a/app/src/main/java/com/github/steroidteam/todolist/model/todo/Task.java
+++ b/app/src/main/java/com/github/steroidteam/todolist/model/todo/Task.java
@@ -11,6 +11,7 @@ import java.util.Objects;
 public class Task {
 
     private String body;
+    private Boolean done;
 
     /**
      * Constructs a new Task with a body.
@@ -23,6 +24,7 @@ public class Task {
             throw new IllegalArgumentException();
         }
         this.body = body;
+        this.done = false;
     }
 
     /**
@@ -54,5 +56,13 @@ public class Task {
         if (o == null || getClass() != o.getClass()) return false;
         Task task = (Task) o;
         return Objects.equals(body, task.body);
+    }
+
+    public Boolean isDone() {
+        return done;
+    }
+
+    public void setDone(Boolean done) {
+        this.done = done;
     }
 }
--- a/app/src/main/java/com/github/steroidteam/todolist/view/adapter/TodoAdapter.java
+++ b/app/src/main/java/com/github/steroidteam/todolist/view/adapter/TodoAdapter.java
@@ -53,9 +53,11 @@ public class TodoAdapter extends RecyclerView.Adapter<TodoAdapter.TaskHolder> {
 
         if (currentTask.isDone()) {
             holder.taskBody.setTextColor(Color.LTGRAY);
+            holder.taskBody.setPaintFlags(holder.taskBody.getPaintFlags() | Paint.STRIKE_THRU_TEXT_FLAG);
             holder.displayDeleteButton();
         } else {
             holder.taskBody.setTextColor(Color.DKGRAY);
+            holder.taskBody.setPaintFlags(0);
             holder.hideDeleteButton();
         }
     }
--- a/app/src/main/java/com/github/steroidteam/todolist/view/adapter/TodoAdapter.java
+++ b/app/src/main/java/com/github/steroidteam/todolist/view/adapter/TodoAdapter.java
@@ -2,25 +2,17 @@ package com.github.steroidteam.todolist.view.adapter;
 
 import android.graphics.Color;
 import android.graphics.Paint;
-import android.text.Editable;
-import android.text.TextWatcher;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.Button;
 import android.widget.CheckBox;
-import android.widget.CompoundButton;
-import android.widget.EditText;
 import android.widget.TextView;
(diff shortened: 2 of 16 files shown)


app/src/main/java/com/github/steroidteam/todolist/model/todo/Task.java
```diff
@@ -11,6 +11,7 @@
 public class Task {
 
     private String body;
+    private Boolean done;
```
shoutizix, line 14: Maybe isDone would be a better name for the boolean

app/src/main/java/com/github/steroidteam/todolist/model/todo/Task.java
```diff
@@ -55,4 +57,12 @@ public boolean equals(Object o) {
         Task task = (Task) o;
         return Objects.equals(body, task.body);
     }
+
+    public Boolean isDone() {
+        return done;
+    }
+
+    public void setDone(Boolean done) {
```
shoutizix, line 65: Same here, maybe "setIsDone" and "getIsDone" would be better for the boolean but maybe it's only a matter of personal taste haha

shoutizix (approved):
Hey,

It looks very good, I've tried it on my phone and the new panel to update the tasks looks great to me too !

As I mentionned the only little thing that maybe you could improve was the name of the boolean so that it starts with a verb but other than that it seems really great nice job !

app/src/main/java/com/github/steroidteam/todolist/view/adapter/TodoAdapter.java
```diff
@@ -56,55 +67,56 @@ public void setTodoList(TodoList todoList) {
...
-    }
-
-    public void setCurrentlyDisplayedUpdateLayoutPos(Integer currentlyDisplayedUpdateLayoutPos) {
-        this.currentlyDisplayedUpdateLayoutPos = currentlyDisplayedUpdateLayoutPos;
-    }
-
     public interface TaskCustomListener {
-        void onClickCustom(int position);
+        void onClickCustom(TaskHolder holder, int position);
+        // void onFocusChangedCustom(int position, boolean hasFocus);
```
VPellenc, line 72: Commented-out code (not sure if intentional)

VPellenc (approved):
Tested it and it looks very nice!

One thing I noticed, when the panel is opened after clicking a task, it is not small on my emulator, it takes all the free space.
There was also the idea of having the Todo go to the bottom when getting marked as done but I think it looks great as is.

theo-damiani:
> 
> 
> Tested it and it looks very nice!
> 
> One thing I noticed, when the panel is opened after clicking a task, it is not small on my emulator, it takes all the free space.
> There was also the idea of having the Todo go to the bottom when getting marked as done but I think it looks great as is.

Thanks, @VPellenc for the review!
Normally the panel looks like this: 


Could you confirm for me that you have the same thing?
This is not a dialog so yes it was built to take a big part of the screen (the same as the recycler view).

Yes, I didn't implement the logic to move a task to the bottom when it is done, because I will prefer to implement this in a PR that takes care of all the sorting aspect of a task as it is quite the same thing.

VPellenc:
> > Tested it and it looks very nice!
> > One thing I noticed, when the panel is opened after clicking a task, it is not small on my emulator, it takes all the free space.
> > There was also the idea of having the Todo go to the bottom when getting marked as done but I think it looks great as is.
> 
> Thanks, @VPellenc for the review!
> Normally the panel looks like this:
> 
> 
> Could you confirm for me that you have the same thing?
> This is not a dialog so yes it was built to take a big part of the screen (the same as the recycler view).
> 
> Yes, I didn't implement the logic to move a task to the bottom when it is done, because I will prefer to implement this in a PR that takes care of all the sorting aspect of a task as it is quite the same thing.

Yes I have the same thing.
It looks a bit weird with the empty space until the Delete button but it's definitely not a blocker.
I could imagine it as an "extension" of an item in the recyclerView, where you can still see the other tasks around but I don't know how hard it would be to implement (if even possible). Either way I think it looks good enough for now.

theo-damiani:
> Yes I have the same thing.
> It looks a bit weird with the empty space until the Delete button but it's definitely not a blocker.
> I could imagine it as an "extension" of an item in the recyclerView, where you can still see the other tasks around but I don't know how hard it would be to implement (if even possible). Either way I think it looks good enough for now.

I designed that way because I thought that we will have space to add more features: like add categories to a task, add a boolean to know if a task is important, ...

theo-damiani:
Is it okay if we merge this PR?

VPellenc:
> > Yes I have the same thing.
> > It looks a bit weird with the empty space until the Delete button but it's definitely not a blocker.
> > I could imagine it as an "extension" of an item in the recyclerView, where you can still see the other tasks around but I don't know how hard it would be to implement (if even possible). Either way I think it looks good enough for now.
> 
> I designed that way because I thought that we will have space to add more features: like add categories to a task, add a boolean to know if a task is important, ...

Yeah I see in that case my alternative makes less sense



> Is it okay if we merge this PR?

Good for me

PR #149, opened 2021-04-14
Feature/recycler view for notes

theo-damiani:
PR so that the note selection uses a recycler view.
No view model yet as the note-database PR #83 has not been merged. But this will not add a lot of work. The bigger part (recycler view and Holder is done).


app/src/main/java/com/github/steroidteam/todolist/view/adapter/NoteAdapter.java
```diff
@@ -0,0 +1,69 @@
...
+    @Override
+    public int getItemCount() {
+        return noteList.size();
+    }
+
+    public UUID getIdOfNote(int index) {
+        return noteList.get(index).getId();
+    }
+
+    public interface NoteCustomListener {
```
sydneyhauke, line 50: Is there no other way than creating its own custom listener ?

sydneyhauke (approved):

app/src/main/java/com/github/steroidteam/todolist/view/adapter/NoteAdapter.java
theo-damiani: I think but this way I can add other methods in the future and it was easier to recover in the activity the Holder of a note.
But sure if you prefer I can maybe look to change this.

shoutizix (approved):
Nice !

This works perfectly on my phone ! The cardview for the notes is to display the images even in the selection am I right ?

I think this would be great to merge it pretty fast, because everytime we'll make modification on the notes part we'll need to adapt this PR to not break code, so the sooner the better I think.

theo-damiani:
> 
> 
> Nice !
> 
> This works perfectly on my phone ! The cardview for the notes is to display the images even in the selection am I right ?
> 
> I think this would be great to merge it pretty fast, because everytime we'll make modification on the notes part we'll need to adapt this PR to not break code, so the sooner the better I think.

Thanks!
This the idea but as there isn't the viewmodel I just put the logo in every notes.

Issue #55, opened 2021-03-19
Create a visual design style


YagoGG:

YagoGG:
![mockups](https://user-images.githubusercontent.com/7356565/112529622-c76f1c00-8da5-11eb-8f94-2b95e4931b0d.png)

Key colors:
- `#F1F1F1` (light grey)
-  `#C8C8C8` (darker grey)
- `#035AA6` (dark blue)
- `#000000` (black)

PR #150, opened 2021-04-15
Allow to attaching the main image (header) to a note

VPellenc:
In this PR, you can add an image in a note by clicking the camera icon in the top right of the header. This will open a file picker, and you can choose the image you want.
In order to add an image in the Android emulator files, you can drag and drop the image in the emulator. I recommend to then move it in the Pictures folder in the SD card as I think it should be the main use of the feature.

I also added methods in the VolatileDatabase that should handle it, although it is not used for now.

However I am not sure how I can test it with Espresso. The closest helper I could find is here https://proandroiddev.com/testing-camera-and-galley-intents-with-espresso-218eb9f59da9 but it's in Kotlin so I will try to make it work at some point later.


app/src/main/java/com/github/steroidteam/todolist/view/NoteDisplayActivity.java
```diff
@@ -21,7 +29,7 @@ protected void onCreate(Bundle savedInstanceState) {
         backButton.setOnClickListener((view) -> finish());
 
         Intent intent = getIntent();
-        UUID id = UUID.fromString(intent.getStringExtra(NoteSelectionActivity.EXTRA_NOTE_ID));
+        // UUID id = UUID.fromString(intent.getStringExtra(NoteSelectionActivity.EXTRA_NOTE_ID));
```
theo-damiani, line 32: commented code

app/src/main/java/com/github/steroidteam/todolist/database/FirebaseDatabase.java
```diff
@@ -146,4 +146,12 @@ public Task getTask(@NonNull UUID todoListID, @NonNull Integer taskIndex)
             throw new DatabaseException(e.toString());
         }
     }
+
+    @Override
+    public void assignMainImage(UUID noteID, String path) {}
+
+    @Override
+    public String getMainImagePath(UUID noteID) {
+        return null;
```
theo-damiani, line 155: maybe you can return `throw new UnsupportedOperationException("TODO");`
So we don't forget to implement it

theo-damiani (approved):
Great work!
Tested and looks fine!

Just one little thing to mention is that we lose the rounded corner when we add an image:

app/src/main/java/com/github/steroidteam/todolist/database/VolatileDatabase.java
```diff
@@ -131,4 +133,32 @@ public void doneTask(UUID todoListID, Integer taskIndex, boolean isDone) {
...
+    }
+
+    /**
+     * Gets the main image of a note.
+     *
+     * @param noteID The id of the note.
+     * @return The path of the image in the SD card, or null if the ID isn't in the database.
+     */
+    @Override
+    public String getMainImagePath(UUID noteID) {
```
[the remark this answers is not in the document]
VPellenc: This makes a lot of sense as well. I don't know what is best but I think I will change it.

VPellenc:
I think it is ready to be merged.

I removed the VolatileDatabase implementation because we will soon no longer use it and because if we want to get a BitmapDrawable from it, it complicates the implementation a lot (a Context is needed in order to access the filesystem, so it would need to be added either in the constructor or in the get method)

theo-damiani:
Hey @VPellenc,
Great job! This is perfect for me, ready to be merged!

Just do you think is it possible to keep the rounded corner of the note after adding an image?

VPellenc:
Oh I forgot about it.
I just pushed a commit that does that, the corner radius might differ a little bit however.

app/src/main/java/com/github/steroidteam/todolist/view/NoteDisplayActivity.java
```diff
@@ -34,4 +42,31 @@ protected void onCreate(Bundle savedInstanceState) {
...
+        if (requestCode == PICK_IMAGE) {
+            Uri uri = data.getData();
+            ConstraintLayout header = findViewById(R.id.note_header);
+            Bitmap bitmap = null;
+            try {
+                InputStream is = getContentResolver().openInputStream(uri);
+                bitmap = BitmapFactory.decodeStream(is);
+                is.close();
+            } catch (Exception e) {
+            }
```
YagoGG, line 65: It might be worth adding a `Toast` here indicating that something went wrong.

app/src/main/java/com/github/steroidteam/todolist/view/NoteDisplayActivity.java
```diff
@@ -34,4 +42,31 @@ protected void onCreate(Bundle savedInstanceState) {
...
+            RoundedBitmapDrawable ob = RoundedBitmapDrawableFactory.create(getResources(), bitmap);
+            ob.setCornerRadius(70);
```
YagoGG, line 67: We are using 25dp-radius corners in the rest of the UI, can we keep it here for consistency? I don't know if unit conversion might be needed (not sure about the units of that 70).

app/src/androidTest/java/com/github/steroidteam/todolist/NoteDisplayActivityTest.java
```diff
@@ -0,0 +1,84 @@
...
+        intending(hasAction(Intent.ACTION_CHOOSER)).respondWith(result);
+    }
+
+    @Test
+    public void filePickingWorks() {
+        onView(withId(R.id.camera_button)).perform(click());
+    }
+
+    private void savePickedImage() {
+        Bitmap bm =
```
YagoGG, line 54: It's not super clear why are we using this image. I guess that the idea is to reuse something that we already have lying around the resources, but it might be better to explicitly add something with a more likely size (e.g. a 1MB picture), or to generate it procedurally as explained below.

YagoGG (changes requested):
Hey, @VPellenc! Thanks a lot for the PR.

The feature looks pretty sweet, although I feel like the following points should be addressed before merging:

1. As @theo-damiani pointed out, the corners should be rounded, but only the bottom ones. Please check out the mockups in #55 for reference.
2. The app crashes if you hit the "Back" button when adding a new image. Let me know if you need more details on the steps to reproduce.

I also left a few inline comments, I hope you find them useful.

Cheers!

sydneyhauke (approved):

app/src/main/res/values/strings.xml
```diff
@@ -20,6 +20,7 @@
     <string name="note_activity_title">My notes</string>
     <string name="search_btn_description">Search</string>
     <string name="back_btn_description">Back</string>
+    <string name="camera_btn_description">Photo</string>
```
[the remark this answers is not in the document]
VPellenc: Oh I didn't know that, makes sense

VPellenc:
> 1. As @theo-damiani pointed out, the corners should be rounded, but only the bottom ones. Please check out the mockups in #55 for reference.

Yeah, I missed it because the image I used for testing had a background very similar to the header in these spots.
I didn't find anything from the standard Android library that does it, so for now I removed the feature and am back to square corners. Since an external library will probably be necessary for this, I think we should push this feature for another sprint. It also checks the non trivial part of sensor usage of the app requirement. https://stackoverflow.com/questions/48669327/individual-rounded-corners-not-all-corners
Of course notify me if I missed something.



> 2\. The app crashes if you hit the "Back" button when adding a new image. Let me know if you need more details on the steps to reproduce.
Yeah I can reproduce. I'm looking into it now.

app/src/main/java/com/github/steroidteam/todolist/view/NoteDisplayActivity.java
VPellenc: Indeed it is possible by using `getResources().getDisplayMetrics().density`

theo-damiani:
> 
> 
> > 1. As @theo-damiani pointed out, the corners should be rounded, but only the bottom ones. Please check out the mockups in #55 for reference.
> 
> Yeah, I missed it because the image I used for testing had a background very similar to the header in these spots.
> I didn't find anything from the standard Android library that does it, so for now I removed the feature and am back to square corners. Since an external library will probably be necessary for this, I think we should push this feature for another sprint. It also checks the non trivial part of sensor usage of the app requirement. https://stackoverflow.com/questions/48669327/individual-rounded-corners-not-all-corners
> Of course notify me if I missed something.
> 

No problem! It was a small detail.
For me, we can merge this PR!

YagoGG (approved):
I just tested the new changes and it looks great. Thanks for the brilliant work, @VPellenc!

PR #151, opened 2021-04-15
Feature/97 database persistency

sydneyhauke:
Persists tasks, notes and todolists in remote database.

The database interface needed to be modified so that calls become asynchronous. This eases the use of the database.


sydneyhauke:
I should also add that I removed the VolatileDatabase. My reasoning is that now we can persist almost anything in the database. What do you think ?

theo-damiani:
> 
> 
> I should also add that I removed the VolatileDatabase. My reasoning is that now we can persist almost anything in the database. What do you think?

Yes for me we can remove the VolatileDatabase. But this will not break the testing part of some activities if we made calls to Firebase?

app/src/main/java/com/github/steroidteam/todolist/view/ListSelectionActivity.java
```diff
@@ -67,6 +85,14 @@ public void openNotes(View view) {
         startActivity(noteSelectionActivity);
     }
 
+    public void createList(View view) {
+        TodoList todoList = new TodoList("New TodoList");
```
YagoGG, line 94: Can we move this to its own `@string`?

app/src/main/res/layout/activity_note_selection.xml
```diff
@@ -66,4 +66,20 @@
 
     </ListView>
 
+    <com.google.android.material.floatingactionbutton.FloatingActionButton
+        android:id="@+id/create_note_button2"
```
YagoGG, line 70: Why the "2" in `create_note_button2`?

sydneyhauke:
> > I should also add that I removed the VolatileDatabase. My reasoning is that now we can persist almost anything in the database. What do you think?
> 
> Yes for me we can remove the VolatileDatabase. But this will not break the testing part of some activities if we made calls to Firebase?
> 
> EDIT: maybe we need to keep it so we don't have any problem with the others PR like for #150

We could, but that would imply additional work on this PR. The Database interface has changed a lot. So I would need to both adapt the VolatileDatabase so that it correctly implements the new interface and add additional methods that I also had to add to fully support all operations needed on the database.
Otherwise, it's the task of the one responsible for #150 to adapt the Database interface so that it correctly implments what he implemented on the VolatileDatabase. Do you see what I mean ?

YagoGG (changes requested):
Hey there, @sydneyhauke! Thanks a bunch for the work you've put in this. I am suuuper hyped about this feature. :rocket: 

I have read the code and it's super clear and easy to read, congrats!

As for the feature, I have tested it in my local environment and it works very well. I found a couple issues, though:

- I cannot open todo lists after creating them. I have checked the file created in Firebase and it looks fine (just an empty list), so I'm not sure what might be going on.
- I managed to create a mpte with empty title. The steps to reproduce are:
    1. Create a note.
    2. Open it, and hit the "back" button real quick (before the note is loaded).
    3. Go back to the "My lists" view.
    4. Go back to the "My notes" view (so we reload the list of notes). _Note: this also makes me think that we should reload the note list when returning from the NoteDisplayActivity`._
    5. As you will see, a new note has appeared with empty title: 
empty note

If your reflexes are not fast enough, you might need to slow down your internet connection. I'm finally putting the crappiness of my room's internet connection to good use!

In any case I am happy with merging this once we have fixed the todo lists, and just filing an issue for now with the "empty note" problem. Let me know your thoughts, and congrats again for the work!

hype overload gif

YagoGG:
As for the `VolatileDatabase`, I am still down for getting rid of it. If other PRs use it they will have to adapt their code (or this PR will have to adapt the codebase, depending on what gets merged first).

sydneyhauke:
Oh god. I just realized that I might have to adapt every Activity test that might end up communicating with the database... No problem before because we had the VolatileDatabase. But now... I have to mock the database since we are now connected to a real remote database. Sh**.

theo-damiani (approved):
Again a big PR!

I have nothing to say about the code I think is very clean!

I tested the app and it works well, I just had an issue:
If you create a todo or note, then you enter it and after you go back to the previous activity, there is a new todo or note that adds up. If you reload the app it disappears so I think it is not a problem with the Firebase database but with the list view used to display the todo/note. I think this error will disappear when we will add the recycler and ViewModel for the selection of a note/todo.

I also ran the test locally and it was successful.

YagoGG (changes requested):
Hey @sydneyhauke, thanks a bunch for the changes!

I just made a second batch of functional tests and the todo lists work like a charm.

However, I cannot see the body of the notes I edit. Here are the steps to reproduce:

1. Create a note.
2. Add any text in its body.
3. Hit the green button to save the note.
4. Hit the "back" button to return to the note list.
5. Re-open the note.
6. There is no body.

The expected behaviour would be to see the text you just wrote in step (2).

Oddly enough, the file is updated in Firebase, so I guess that it's an issue either when the file is downloaded, or when we read its contents to populate the view.

Let me know if you need any further details to reproduce, I'll be glad to oblige!

Please also check out the two inline comments I left in my last review, as they have not been addressed yet.

Cheers!

sydneyhauke:
> Hey @sydneyhauke, thanks a bunch for the changes!
> 
> I just made a second batch of functional tests and the todo lists work like a charm.
> 
> However, I cannot see the body of the notes I edit. Here are the steps to reproduce:
> 
>     1. Create a note.
> 
>     2. Add any text in its body.
> 
>     3. Hit the green button to save the note.
> 
>     4. Hit the "back" button to return to the note list.
> 
>     5. Re-open the note.
> 
>     6. There is no body.
> 
> 
> The expected behaviour would be to see the text you just wrote in step (2).
> 
> Oddly enough, the file is updated in Firebase, so I guess that it's an issue either when the file is downloaded, or when we read its contents to populate the view.
> 
> Let me know if you need any further details to reproduce, I'll be glad to oblige!
> 
> Please also check out the two inline comments I left in my last review, as they have not been addressed yet.
> 
> Cheers!

I think I know why the body isn't saved. My current solution saves the body only if the back button (upper left on the screen) is pressed. I should save the body whenever the activity looses focus.

sydneyhauke:
Actually, @YagoGG, instead of saving the body whenever we click the green button, shouldn't we instead save the body only if the activity looses focus ?

YagoGG:
That's actually my preference, I was going to suggest to remove the green button in the next sprint. So if we can do it now that's awesome :)

YagoGG:
If it's easier for you, we can also do it whenever the `EditText` loses focus. I guess the event handler is also triggered if the activity is finished. Just an idea in case it might make the implementation simpler!

theo-damiani (approved):
Great works!
Looks and works well for me!

Sometimes when you create a note go back to the selection, you can have twice the note but one without a title.
But you know with the ViewModel it will be easier to display correctly the notes. When we will merge this PR I will take care of plugging with the viewmodel for the notes

theo-damiani:
In the database, we have `renameTask`, `setTaskDone`, and maybe more stuff like this that changes an already existing task.
Do you think that it will be easier to have a method updateTask as we have for todo: `updateTodoList`.
And we could have the same for the notes.

sydneyhauke:
> Hey @sydneyhauke, thanks a bunch for the changes!
> 
> I just made a second batch of functional tests and the todo lists work like a charm.
> 
> However, I cannot see the body of the notes I edit. Here are the steps to reproduce:
> 
>     1. Create a note.
> 
>     2. Add any text in its body.
> 
>     3. Hit the green button to save the note.
> 
>     4. Hit the "back" button to return to the note list.
> 
>     5. Re-open the note.
> 
>     6. There is no body.
> 
> 
> The expected behaviour would be to see the text you just wrote in step (2).
> 
> Oddly enough, the file is updated in Firebase, so I guess that it's an issue either when the file is downloaded, or when we read its contents to populate the view.
> 
> Let me know if you need any further details to reproduce, I'll be glad to oblige!
> 
> Please also check out the two inline comments I left in my last review, as they have not been addressed yet.
> 
> Cheers!

Last commits should have fixed the issue you described.

app/src/main/java/com/github/steroidteam/todolist/view/ListSelectionActivity.java
sydneyhauke: Fixed

app/src/main/res/layout/activity_note_selection.xml
sydneyhauke: Fixed

theo-damiani (approved):

app/src/main/java/com/github/steroidteam/todolist/database/FirebaseDatabase.java
```diff
@@ -2,148 +2,222 @@
...
-                    .thenApply(JSONSerializer::serializeTodoList)
-                    .thenApply(
-                            serializedTodoList ->
-                                    serializedTodoList.getBytes(StandardCharsets.UTF_8))
-                    .thenApply(bytes -> this.storageService.upload(bytes, listPath))
-                    .get();
-        } catch (ExecutionException | InterruptedException e) {
-            throw new DatabaseException(e.toString());
-        }
+        return getTodoList(todoListID)
```
YagoGG, line 99: I like this, pretty smart!

app/src/main/java/com/github/steroidteam/todolist/todo/TodoListCollection.java
```diff
@@ -0,0 +1,33 @@
+package com.github.steroidteam.todolist.todo;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.UUID;
+
+public class TodoListCollection {
```
YagoGG, line 7: Why creating this class, instead of using directly an `ArrayList<TodoList>`? Seems like we're basically wrapping the main methods of a normal `ArrayList`.

app/src/main/java/com/github/steroidteam/todolist/view/ListSelectionActivity.java
```diff
@@ -15,30 +15,46 @@
...
         ListView listView = findViewById(R.id.activity_list_selection_itemlist);
         adapter = new todoListAdapter(todoLists);
         setListViewSettings(listView);
+
+        database = DatabaseFactory.getDb();
+
+        database.getTodoListCollection()
+                .thenAccept(
+                        (todoListCollection -> {
+                            for (int i = 0; i < todoListCollection.getSize(); i++) {
```
YagoGG, line 49: If we decide to stick with the `TodoListCollection` (instead of a plain `ArrayList`), it might be worth making it an `Iterable`, for situations like this one. See [this](https://stackoverflow.com/a/21512315) for reference.

YagoGG (approved):
Hey @sydneyhauke! All good now. :+1: 

I left a few minor inline comments in case you want to consider them.

In any case I am very comfortable merging this PR in its current state. Congratulations man, brilliant work! :gem:

Issue #161, opened 2021-04-16
Integrate the location in the notes


shoutizix:
We should integrate the location in the notes class and make the corresponding changes

Issue #147, opened 2021-04-11
Add a location (chosen or current location of the user) in the NoteDisplayActivity


shoutizix:
Right now it only show the current GPS location but there's nos synchronization between the MapActivity and the NoteDisplayActivity. We have to make them communicate to show the location in the NoteDisplayActivity.

PR #169, opened 2021-04-21
Feature/location in notes

shoutizix:
Hey guys,

This PR concern #161 #147 . It use the PR #168 as basis, so maybe I will only use this one to have everything merged.
Anyway, I've made modification so that the location of the last put marker is now displayed in the view of the note. 
I have also added a Latlng attribute for the note and not the name, because with Geocoder we can recover the name of the location by using the "getFromLocation" methode.

### DIFF
--- a/app/src/androidTest/java/com/github/steroidteam/todolist/MapsActivityTest.java
+++ b/app/src/androidTest/java/com/github/steroidteam/todolist/MapsActivityTest.java
@@ -1,7 +1,13 @@
 package com.github.steroidteam.todolist;
 
+import static androidx.test.espresso.Espresso.onView;
+import static androidx.test.espresso.action.ViewActions.click;
+import static androidx.test.espresso.action.ViewActions.pressKey;
+import static androidx.test.espresso.action.ViewActions.typeText;
+import static androidx.test.espresso.matcher.ViewMatchers.withId;
 import static androidx.test.platform.app.InstrumentationRegistry.getInstrumentation;
 
+import android.view.KeyEvent;
 import androidx.test.espresso.intent.Intents;
 import androidx.test.ext.junit.rules.ActivityScenarioRule;
 import androidx.test.ext.junit.runners.AndroidJUnit4;
@@ -54,6 +60,16 @@ public class MapsActivityTest {
                 uiObjectNotFoundException.printStackTrace();
             }
         }
+    }
+
+    @Test
+    public void searchViewWorksCorrectly() {
+        onView(withId(R.id.sv_location))
+                .perform(click(), typeText("Lausanne"), pressKey(KeyEvent.KEYCODE_ENTER));
+        waitFor(4000);
+        UiDevice device = UiDevice.getInstance(getInstrumentation());
+        UiObject marker = device.findObject(new UiSelector().descriptionContains("Lausanne"));
+        Assert.assertNotNull(marker);
     }
 
     private void waitFor(int duration) {
--- a/app/src/androidTest/java/com/github/steroidteam/todolist/MapsActivityTest.java
+++ b/app/src/androidTest/java/com/github/steroidteam/todolist/MapsActivityTest.java
@@ -67,19 +67,23 @@ public class MapsActivityTest {
 
     @Test
     public void searchViewWorksCorrectlyWithCorrectLocation() {
+        String TEST_LOCATION = "Lausanne";
         onView(withId(R.id.sv_location))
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/github/steroidteam/todolist/view/NoteDisplayActivity.java
```diff
@@ -43,6 +46,26 @@ protected void onCreate(Bundle savedInstanceState) {
...
+            TextView locationText = (TextView) findViewById(R.id.note_location);
+            locationText.setText(location);
+        }
+    }
+
+    @Override
+    protected void onActivityResult(int requestCode, int resultCode, Intent data) {
+        super.onActivityResult(requestCode, resultCode, data);
+        if (requestCode == LAUNCH_SECOND_ACTIVITY && resultCode == Activity.RESULT_OK) {
+            String location = data.getStringExtra("nameLocation");
```
theo-damiani, line 66: I think it is better to use private final static attribute for the Extra

app/src/main/java/com/github/steroidteam/todolist/view/MapsActivity.java
```diff
@@ -154,6 +153,20 @@ public void onMapReady(GoogleMap map) {
         getDeviceLocation();
     }
 
+    @Override
+    public void onBackPressed() {
+        Intent returnIntent = new Intent();
+        if (marker != null) {
+            returnIntent.putExtra("location", marker.getPosition());
```
theo-damiani, line 160: same here for the extra. So you know if you don't have to change every file if you have an attribute in the MapsActivity or NoteActivity

theo-damiani (approved):
Great work!

I tested and it worked well! definitely a good addition for the demo Friday!

I just have small remarks:

- when you go to the mapActivity and go back (without searching any location), the location is set to "I'm here" it would be preferable to have "no location" I think.


- It would be great to have a button in the map activity that allows you to go back in the note activity. So you confirm that you want this position.
 Currently, you have nothing instead of the arrow at the bottom (the on of the phone):

app/src/androidTest/java/com/github/steroidteam/todolist/MapsActivityTest.java
```diff
@@ -56,6 +67,36 @@ public void markerIsCorrectlyPlacedAtDefaultLocationOrUserLocation() {
         }
     }
 
+    @Test
+    public void searchViewWorksCorrectlyWithCorrectLocation() {
+        String TEST_LOCATION = "Lausanne";
+        onView(withId(R.id.sv_location))
+                .perform(click(), typeText(TEST_LOCATION), pressKey(KeyEvent.KEYCODE_ENTER));
+        waitFor(4000);
```
sydneyhauke, line 75: Why do you need to wait ?

app/src/androidTest/java/com/github/steroidteam/todolist/NoteDisplayActivityTest.java
```diff
@@ -32,9 +39,31 @@ public void openMapsActivityWorks() {
...
+        onView(withId(R.id.location_button)).perform(click());
 
-        Intents.intended(Matchers.allOf(IntentMatchers.hasComponent(MapsActivity.class.getName())));
+        Intents.intended(Matchers.allOf(hasComponent(MapsActivity.class.getName())));
+        Intents.release();
+    }
+
+    @Test
+    public void locationNameIsCorrectlyUpdated() {
+        Intents.init();
```
sydneyhauke, line 50: You can place these static methods in @Before and @After annotated methods.

app/src/main/java/com/github/steroidteam/todolist/view/MapsActivity.java
```diff
@@ -154,6 +153,20 @@ public void onMapReady(GoogleMap map) {
...
+            returnIntent.putExtra("nameLocation", marker.getTitle());
+        } else {
+            returnIntent.putExtra("location", (LatLng) null);
```
sydneyhauke, line 163: Hardcoded "extra" strings. Better to put them in public final static fields.

app/src/androidTest/java/com/github/steroidteam/todolist/MapsActivityTest.java
shoutizix: Because there is an animation that moves to the searched place, so I wait until the animation have finished to test the marker

sydneyhauke: Perhaps we should deactivate animation during tests...

sydneyhauke (approved):

PR #172, opened 2021-04-22
Feature/internal file system

theo-damiani:
local file system.
I added a LocalDatabase even if in fact it is almost the same as the FirebaseDatabase, only the filesystem changes in it.


theo-damiani:
https://github.com/steroid-team/app/pull/167#issue-616879054

theo-damiani:
I need to mock the local system service but also the context, So I decided to mock the method needed but this doesn't work. 
I tried to mock the local system service and the context but the service doesn't use the mock context.

theo-damiani:
some comments so I don't forget:

- why we use the Id to put a note: `putNote(UUID, Note)`. For todo we did `putTodo(Todo)`
- twice check non null on the user in firebase system service: `Objects.requireNonNull(user);` x2

PR #174, opened 2021-04-22
Feature/color picker

lgraziano:
A color picker for the drawing activity
It is based on my previous branch, drawActivity, so we can focus on this one instead and remove the other.
It won't pass the tests because of some other classes, if someone understands the problem some help is welcome !


theo-damiani:
For those errors,  I think you should just rerun the test : 

com.github.steroidteam.todolist.NoteDisplayActivityTest > openMapsActivityWorks[test(AVD) - 10] �[31mFAILED �[0m
	androidx.test.espresso.PerformException: Error performing 'single click - At Coordinates: 159, 123 and precision: 16, 16' on view 'with id is <com.github.steroidteam.todolist:id/note_header>'.
	at androidx.test.espresso.PerformException$Builder.build(PerformException.java:5)

com.github.steroidteam.todolist.NoteSelectionActivityTest > openListWorks[test(AVD) - 10] �[31mFAILED �[0m
	java.lang.IllegalStateException: #init was called twice in a row. Make sure to call #release after every #init
	at androidx.test.espresso.intent.Checks.checkState(Checks.java:70)

theo-damiani (approved):
Hey @lgraziano, great work!
Tested and work perfectly! I didn't face any issues.

I just have some remarks about the design:
- When you pick a new color, the black circle doesn't change:

- It would be great to have a button to return to the previous activity

This PR is great and definitely a good feature to show tomorrow.
But for the next spring, I think it would be great to continue to improve the design of the activity, add tests that actually test, and maybe have a first look at how we could store a drawing in the database.

lgraziano:
Thanks @theo-damiani for the review ! I just applied the fixes

app/src/main/res/layout/fragment_drawing.xml
```diff
@@ -0,0 +1,148 @@
...
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:foregroundTint="#FFFFFF"
+    tools:context=".view.DrawingFragment">
+
+    <Button
+        android:id="@+id/erase_button"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:text="Erase"
```
shoutizix, line 14: maybe put the Erase in the string.xml file so it's not hardcoded

shoutizix (approved):
Hey,

Nice, it works like a charm ! I think it's a really great new feature to have with sensor usage and we'll can show it tomorrow that's perfect !

Issue #140, opened 2021-04-02
Create a view for recording voice memos within a note


sydneyhauke:
Features that we expect:

- Start recording the memo.
- Stop the recording.
- Pause the recording (so that it can be resumed). Maybe this can be merged with the previous feature.
- Replay the recording so far.
- Close the recording view.

It would also be great if the user could move around the recorded memo while recording (go forward and backward, like when you're playing a song).

YagoGG:
@sydneyhauke you might be interested in checking this out: https://github.com/3llomi/RecordView

Maybe it's not what we're looking for, but there might be other interesting options available.

PR #190, opened 2021-04-28
Voice recording activity

shoutizix:
It addresses #140.
It simply add an activity for the audio recorder and connect it with the note display activity. For now it only record on one hardcoded file, and it's not saved with the note. I also have added a media player to play the audio.


lgraziano (approved):
Very nice ! It works on my device, but the sound is kinda weak. Except for that it is just fine 👏

app/src/androidTest/java/com/github/steroidteam/todolist/AudioRecorderFragmentTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+    @Test
+    public void toastAndTextWhenRecordingShowsCorrectly() {
+        onView(withId(R.id.record_button)).perform(click());
+        onView(withText(R.string.is_recording))
+                .inRoot(new ToastMatcher())
+                .check(matches(isDisplayed()));
+        onView(withId(R.id.record_text)).check(matches(withText(R.string.stop_record_button)));
+        // Wait to avoid error with Toast messages that is still displayed
+        waitFor(3000);
```
theo-damiani, line 35: Is this wait necessary if you do nothing after in the test ?

theo-damiani (approved):
Great PR!
This works fine, for me the sound wasn't so much low but kind of 'blurry' after it is maybe just my microphone or the quality of the recording, I think you can't do so much about it.
For me it is ok to merged it.

app/src/androidTest/java/com/github/steroidteam/todolist/AudioRecorderFragmentTest.java
shoutizix: I put this to avoid that the toast is still displayed for the next test so I think this is useful

PR #191, opened 2021-04-28
feature: Implement rich editor for notes

### LINKED ISSUE #142: Allow embedding images in the text of the notes

YagoGG:
A proper editor that allows formatting the text and embedding images.

Closes #142.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -49,6 +49,7 @@ dependencies {
     implementation 'androidx.lifecycle:lifecycle-extensions:2.2.0'
     implementation 'com.google.android.gms:play-services-maps:17.0.0'
     implementation 'com.google.android.gms:play-services-location:18.0.0'
+    implementation 'jp.wasabeef:richeditor-android:2.0.0'
     testImplementation 'junit:junit:4.13.2'
     testImplementation 'org.mockito:mockito-core:3.9.0'
     androidTestImplementation 'org.mockito:mockito-android:3.9.0'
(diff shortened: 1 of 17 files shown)


shoutizix (commented):
Hey @YagoGG ,
This is awesome and really nice designed nice work !

I just have a problem with the "quotation" or "tabulation" feature, maybe it's my bad but I can't write a ligne at the original pos, so without tabulation after writing some text with tabulation.
For example : 
InkedScreenshot_20210504-225521_LI
Here I can't write back at the left of the screen as the first line.
Tell me if I'm not clear enough I'll make a video.

But all the other things seems to work perfectly !

YagoGG:
Hey, @shoutizix! Thanks a lot for the review.

You are absolutely right, I hadn't realised!

After doing some research, apparently it's an issue in the component used for the editor (`RichEditor`). I have checked out if we could fix the issue and send a PR upstream, but unfortunately it looks like a non-trivial thing and it could take some time.

Since blockquotes are not a critical part of the feature, I decided to remove them for now. Let me know if that's a problem, and we can try to discuss alternatives.

Cheers!

theo-damiani (approved):
Tested and look awesome!
Really I have nothing to said it's super clean!

Maybe just check the lint warnings about your Hardcoded string in the items of the rich editor.
And also I find that it is a little hard to add text after an image:

For example here, I need to click where it is yellow to get the cursor after the image and then add a new line and write my text. I don't know if it is possible that when you add an image, just after a new line is created so you can scroll down and click on it to add text. 
I mean having this:

But you it is realy one detail.
The PR is really great!

shoutizix (approved):
I totally agree with you @YagoGG it's better to remove it now and if we have time maybe add it back.

So of course right now we can merge this awesome PR nice !

PR #204, opened 2021-05-01
FIX: double tap to set a task has done

theo-damiani:


shoutizix (approved):
Nice work !

It seems indeed better to use the thenCompose and thenAccept here to use benefits from Futures and it fixes the issue of the double tap so it's amazing !

We can definitely merged it.

PR #209, opened 2021-05-03
Feature/note selection mvvm

theo-damiani:
The PR include ViewModel for note selection, and the following functionalities: creating, deleting and renaming note.
I still need to write test

### DIFF
--- a/app/src/main/java/com/github/steroidteam/todolist/view/NoteSelectionFragment.java
+++ b/app/src/main/java/com/github/steroidteam/todolist/view/NoteSelectionFragment.java
@@ -8,6 +8,7 @@ import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.EditText;
+import android.widget.Toast;
 
 import androidx.annotation.NonNull;
 import androidx.appcompat.view.ContextThemeWrapper;
@@ -29,7 +30,7 @@ import java.util.UUID;
 public class NoteSelectionFragment extends Fragment {
 
     public static final String NOTE_ID_KEY = "id";
-    ArrayList<Note> notes;
+
     private NoteSelectionViewModel viewModel;
     private NoteArrayListAdapter adapter;
 
@@ -53,7 +54,9 @@ public class NoteSelectionFragment extends Fragment {
                 .observe(
                         getActivity(),
                         (noteList) -> {
+                            System.err.println("ONAAAAAAAAAAAAAAAAAAAAAAAA " + noteList.toString());
                             adapter.setNoteList(noteList);
+                            adapter.notifyDataSetChanged();
                         });
 
         return root;
@@ -67,13 +70,18 @@ public class NoteSelectionFragment extends Fragment {
                 bundle.putSerializable(NOTE_ID_KEY, holder.getId());
                 Navigation.findNavController(holder.itemView).navigate(R.id.nav_note_display, bundle);
             }
-        }
+
+            @Override
+            public void onNoteDelete(UUID id) {
+                removeNote(id);
+            }
+        };
     }
 
-    public void createNote(View view) {
+    private void createNote(View view) {
         Context context = new ContextThemeWrapper(getActivity(), R.style.Dialog);
         MaterialAlertDialogBuilder builder = new MaterialAlertDialogBuilder(context);
-        builder.setTitle(getString(R.string.add_todo_suggestion));
+        builder.setTitle(getString(R.string.new_note_btn_description));
 
         LayoutInflater inflater = this.getLayoutInflater();
         View user_input = inflater.inflate(R.layout.alert_dialog_input, null);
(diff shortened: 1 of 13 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -16,6 +16,9 @@
     <string name="undo">UNDO</string>
     <string name="delete">DELETE</string>
     <string name="save">SAVE</string>
```
YagoGG, line 18: I know these aren't modified by your PR, but would you mind fixing the capitalisation (e.g. _Undo_ instead of _UNDO_)? I don't know why are they uppercase in the first place, but makes the "SAVE" button look weird when renaming a note.

app/src/main/res/values/strings.xml
```diff
@@ -29,4 +32,7 @@
     <string name="menu_notes">Notes</string>
     <string name="menu_todo_lists">Todos</string>
     <string name="menu_btn_description">Open menu</string>
+
+    <string name="delete_note_suggestion">Are you sure to delete this note ?</string>
+    <string name="rename_note_suggestion">Enter a new title for your Note:</string>
```
YagoGG, line 37: `Note` should be all lowercase.

app/src/main/res/values/strings.xml
```diff
@@ -29,4 +32,7 @@
     <string name="menu_notes">Notes</string>
     <string name="menu_todo_lists">Todos</string>
     <string name="menu_btn_description">Open menu</string>
+
+    <string name="delete_note_suggestion">Are you sure to delete this note ?</string>
```
YagoGG, line 36: Space before question mark.

YagoGG (changes requested):
Hey @theo-damiani! Thanks a lot for the PR, it looks really sweet!

Could we add a little margin to the sides of the text input? It feels weird that the "Enter a new title..." text is properly padded, but the input right below isn't.
renaming dialog

As well regarding the dialogs, they appear behind the dimmed background (just like in #119).

I also like a lot the swipe to rename/delete method in that PR. I feel like it would be good for consistency to have it in both the `ListSelectionFragment` and the `NoteSelectionFragment`, but I am not sure if it would take a lot of time to implement so it's totally up to you.

Cheers!

app/src/main/java/com/github/steroidteam/todolist/view/NoteSelectionFragment.java
```diff
@@ -35,43 +39,119 @@ public View onCreateView(
...
+            }
+
+            @Override
+            public void onLongClickCustom(Note note) {
+                renameNote(note);
+            }
+        };
+    }
+
+    private void createNote(View view) {
```
sydneyhauke, line 80: I think it needs some factorization with removeNote and renameNote. They have a lot of duplicate code

app/src/main/java/com/github/steroidteam/todolist/viewmodel/NoteSelectionViewModel.java
```diff
@@ -0,0 +1,67 @@
...
+        ArrayList<Note> privateArrayList = new ArrayList<>();
+        if (noteIDList.size() == 0) {
+            noteList.setValue(privateArrayList);
+        } else {
+            for (int i = 0; i < noteIDList.size(); i++) {
+                this.database
+                        .getNote(noteIDList.get(i))
+                        .thenAccept(
+                                note -> {
+                                    privateArrayList.add(note);
```
sydneyhauke, line 39: I wonder if this is thread-safe. Notes are fetched concurrently and are added out-of-order in the list. Maybe something to address later.

sydneyhauke (changes requested):
Great job ! 

I suggest refactoring some code first (see my comment) and then we can merge.

sydneyhauke (approved):

app/src/androidTest/java/com/github/steroidteam/todolist/CustomMatchers.java
```diff
@@ -0,0 +1,90 @@
...
+                return "wait for at least " + millis + " millis.";
+            }
+
+            @Override
+            public void perform(final UiController uiController, final View view) {
+                uiController.loopMainThreadUntilIdle();
+                uiController.loopMainThreadForAtLeast(millis);
+            }
+        };
+    }
```
YagoGG: We have a separate directory for custom `ViewAction`s, see `com.github.steroidteam.todolist.customviewactions`.

YagoGG: Maybe you can add the custom `ViewAction` as a class there, and then create `waitAtLeastHelper` as a static method.

YagoGG: In fact, I just realized that this helper is not used at all, so I am going to remove it in an additional commit.

YagoGG (approved):
Hey @theo-damiani! It looks perfect now, thanks a lot!

I pushed a couple minor commits, but this is ready to go. Cheers!

PR #211, opened 2021-05-05
Notification date

shoutizix:
This prepare our code to have a notification based on a due date.


app/src/main/java/com/github/steroidteam/todolist/view/ItemViewFragment.java
```diff
@@ -137,4 +149,32 @@ public void openUpdateLayout(TodoAdapter.TaskHolder holder, final int position)
     public void checkBoxTaskListener(final int position, final boolean isChecked) {
         itemViewModel.setTaskDone(position, isChecked);
     }
+
+    public void createNotificationChannel() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            CharSequence name = "ReminderChannel";
+            String description = "Channel for the reminder";
```
theo-damiani, line 156: Hardcoded string

theo-damiani (changes requested):
Hey @shoutizix,
Nice work!

It is all good for me.
I am just wondering if it is possible to create maybe a class `NotificationChannel` maybe in the broadcast package that you created. I mean, it will be great that the `createNotificationChannel` and `createNotification` can be re-used in other fragments. For example, if we want to add reminders for Note in the future. Of course if it is possible.

theo-damiani (approved):
Awesome thanks for the changes!

shoutizix:
Thanks a lot for your review @theo-damiani and yes I think you are right so I've made the appropriate changes !

PR #212, opened 2021-05-05
feature: Add support for dates in Task objects

### LINKED ISSUE #178: Modify Task model so that it can hold a reminder

YagoGG:
An initial effort to store due dates in each `Task`, plus some housekeeping.

Closes #178.

### DIFF
--- a/app/src/main/java/com/github/steroidteam/todolist/model/todo/Task.java
+++ b/app/src/main/java/com/github/steroidteam/todolist/model/todo/Task.java
@@ -47,7 +47,13 @@ public class Task {
 
     @Override
     public String toString() {
-        return "Task{" + "body='" + body + "'" + "}";
+        return "Task{"
+                + "body='"
+                + body
+                + "', "
+                + "done="
+                + done
+                + "}";
     }
 
     @Override
--- a/app/src/main/java/com/github/steroidteam/todolist/model/todo/Task.java
+++ b/app/src/main/java/com/github/steroidteam/todolist/model/todo/Task.java
@@ -52,7 +52,7 @@ public class Task {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
         Task task = (Task) o;
-        return Objects.equals(body, task.body);
+        return Objects.equals(body, task.body) && Objects.equals(done, task.isDone());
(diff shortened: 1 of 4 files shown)


theo-damiani (approved):
Thanks for those changes

app/src/main/java/com/github/steroidteam/todolist/model/todo/Task.java
```diff
@@ -65,4 +70,17 @@ public Boolean isDone() {
...
+    public void setDueDate(Date dueDate) {
+        Objects.requireNonNull(dueDate);
+        this.dueDate = dueDate;
+    }
+
+    public void removeDueDate() {
+        this.dueDate = null;
+    }
+
+    public Date getDueDate() {
```
sydneyhauke, line 83: I would suggest to go "functionnal". Why not return an Optional<Date> since the dueDate can be null ? This forces us to deal with both cases and makes our code less error-prone...

sydneyhauke (changes requested):

app/src/main/java/com/github/steroidteam/todolist/model/todo/Task.java
YagoGG: It's just a matter of consistency; as you can see the `done` attribute could be `null` as well but we do not do such a thing.

I agree that it could be useful to do such a thing, but in that case I believe we should apply the same pattern to all of our models in a task of their own (because otherwise we are going to have inconsistent APIs).

What do you think?

sydneyhauke: You're right. In the current state of the project, maybe this issue is out of the scope of this PR. If the rest of the team agrees, we could create a new refactoring task in an upcoming sprint to rework the way we handle these details. I just want to make sure we don't create technical dept and make our development harder in the future whenever some null value is not handled correctly.

sydneyhauke (approved):

sydneyhauke:
In that case, I will merge this PR right now since every change has been approved.

YagoGG:
Merged, thanks a lot @theo-damiani and @sydneyhauke for the reviews!

PR #213, opened 2021-05-05
Remove artifacts from cirrus report

theo-damiani:
Yago added so I could debug a test.
But I don't know if you want to keep it.


YagoGG:
Since it does not do any harm and it could be useful in the future, I'd say that we can keep it there.

@theo-damiani please feel free to reopen if you think otherwise! :)

PR #217, opened 2021-05-06
add save drawing function

lgraziano:


app/src/main/java/com/github/steroidteam/todolist/view/DrawingFragment.java
```diff
@@ -131,6 +139,24 @@ public void applyColorButton(View view) {
         drawingCanvas.setVisibility(View.VISIBLE);
     }
 
+    public void saveButton(View view) {
+        Date date = new Date();
+        SimpleDateFormat sdf = new SimpleDateFormat("yyyyMMddHHmmss");
+        String fileName = "bitmap" + sdf.format(date) + ".jpg";
```
theo-damiani, line 145: Maybe it is easier to use `UUID.randomUUID()`

app/src/main/res/layout/fragment_drawing.xml
```diff
@@ -55,6 +55,21 @@
         app:layout_constraintStart_toStartOf="parent"
         app:layout_constraintTop_toBottomOf="@+id/drawSpace" />
 
+    <Button
+        android:id="@+id/save_button"
```
theo-damiani, line 59: maybe you could precise this id.
for example: `"@+id/darwing_fragment_save_button"`

theo-damiani (approved):
Hey @lgraziano,
Great works! 

I just get a problem when running your code:
the drawing isn't displayed in the rich editor as I get this:
aa
But you showed us that it was working, so I think I just didn't give all the permission to my emulator.

shoutizix:
Hey @lgraziano !
I've ran your code on emulator and I have the same error that Theo had. I've also ran it on my phone and I can press on the button save but it just does nothing as if it wasn't saved at all.

theo-damiani:
The error I have with the automatic embedding:

app/src/main/java/com/github/steroidteam/todolist/view/DrawingFragment.java
```diff
@@ -46,40 +53,46 @@ public View onCreateView(
...
         drawingCanvas = new DrawingView(getContext());
         canvasLayout.addView(drawingCanvas);
         colorPicker.addSaturationBar(saturationBar);
         colorPicker.addValueBar(valueBar);
 
         setUpButtonAndColor(root);
+        // if (ContextCompat.checkSelfPermission(getContext(), android.permission.) !=
+        // PackageManager.PERMISSION_GRANTED){
+
+        // }
```
YagoGG, line 84: I guess we can remove this?

YagoGG (approved):
Hey @lgraziano! Thanks a bunch for the work.

I have taken a look at the code and it looks really good. After the clarification on how to use it, I have been able to reproduce the feature as well and it works like a charm.

I'd suggest that maybe we can create a follow up issue, so we can add the drawing to the note's body at some other point. But I am more than happy to have this feature merged now! :smile: 

I am pushing just a minor commit removing commented out code. The rest is perfect.

Cheers!

app/src/main/java/com/github/steroidteam/todolist/view/DrawingFragment.java
lgraziano: Thank you very much ! Absolutely, I should have removed it but I missed it 👍

Issue #154, opened 2021-04-16
Store voice memos in the database


YagoGG:
This is the DB backend for the voice memo feature.

PR #218, opened 2021-05-06
Feature/persist audio memo

sydneyhauke:
Addresses issue #154.

Audio memos can now be stored in the database and fetched back.

A few drawbacks to highlight :

1. It handles only one audio memo per Note
2. There is no caching. Every time we open a voice memo, it is fetched from the database.
3. The synchronisation with the database might not be handled correctly with the regard to the fragment's lifecycle. For example, when we loose focus from the audio recording fragment, an uploading task of the voice memo is issued and performed in the background. This task is left alone and not handled anymore.

Tests remain to be written.


theo-damiani (approved):
Great works!

I saw no problems with your PR. 
Your code helps me a lot with my PR about note header persistence.
So thanks for the great work!

Just don't forget to run spotless

shoutizix (approved):
Tested on my phone and it works great that's really nice !
All the code seems great too so I think we can definitely merge this PR now !
Nice work

PR #236, opened 2021-05-07
tests: Grant required permissions within test suite

YagoGG:
Sometimes the app asks for permisssions when testing. I have not been able to
deterministically reproduce when that happens, but these changes should help
keep it covered.


shoutizix (approved):
Great ! I've also added this rules for the test I've made for Location Notification PR #238

lgraziano (approved):
Very usefeul, thanks !
@theo-damiani if I remember correctly you also encountered a problem with the file access permission when saving a drawing ? If that's the case I guess we could use the same solution

PR #238, opened 2021-05-08
Location notification

shoutizix:
Reminder notification based on the location given by the user
It addresses #183


shoutizix:
I can't manage to do the test better because I don't have access to the location of the emulator of Cirrus.

theo-damiani (approved):
Hey Noah!
Nice works.
If I understood well, this is the PR to set all we need to have notifications with the location. 
But the layout to add a location to a task isn't implemented yet.
So yes nothing particular to say. Maybe I'm not sure that your test is actually testing something as you commented assertion.
Because in the ItemViewFragment, I don't see anywhere a toast or a way to add a location. But yes you just split your task with several PR and this is very good!

sydneyhauke (approved):

PR #245, opened 2021-05-13
feature: Allow exporting tasks to an external calendar

YagoGG:
This new feature allows exporting a task to an external calendar.

If a due date has been set for the task, it will automatically appear as the start date. If not, the current one will be used.


theo-damiani (approved):
Let's merge it! 👍👍

lgraziano (approved):
Nice, great job !

PR #247, opened 2021-05-13
Location reminder in task

sydneyhauke:
Addresses #227.

For now it justs adds a (nullable) Location field in the Task class.

I would like your input on what I could further add.


app/src/main/java/com/github/steroidteam/todolist/model/todo/Task.java
```diff
@@ -39,6 +42,15 @@ public String getBody() {
         return this.body;
     }
 
+    public void setLocationReminder(Location location) {
+        Objects.requireNonNull(location);
```
YagoGG, line 46: If we add this condition (which I think is good, just like we do with due dates), we need another method to remove the location reminder. See `Task.removeDueDate` for reference.

YagoGG (changes requested):
Hey, @sydneyhauke! Thanks a bunch for this.

I am missing a method (see inline comment), as well as the corresponding modifications to the `JSONSerializerTest`. See #249 for context, and #212 as an example.

Cheers!

sydneyhauke:
Ok, so I decided to use this already existing branch for issue #247. Now it it possible to add a location reminder in a task by pinpointing on a map. The location remains persistent in the database.

sydneyhauke:
Ok, so I decided to use this already existing branch for issue #247. Now it it possible to add a location reminder in a task by pinpointing on a map. The location remains persistent in the database.

theo-damiani (approved):
Nice PR, very good work!

For me, we can merge it, maybe just remove the TodoReposiroty and ItemViewModel file.

YagoGG (approved):
Thanks a lot for the work, @sydneyhauke!

ship it!

Issue #237, opened 2021-05-07
Task update layout is transparent to clicks


YagoGG:
# Steps to reproduce

1. Open the app & log in.
2. Create a new todo list with any name.
3. Enter the new todo list.
4. Create 3 different tasks, with any body. 
![3 tasks](https://user-images.githubusercontent.com/7356565/117506337-ba893f00-af85-11eb-9b3b-2b1c34c91d43.png)
5. Tap the first task to open the edition layout. 
![edition layout](https://user-images.githubusercontent.com/7356565/117506493-f91ef980-af85-11eb-8618-206141296993.png)
6. Instead of the task that is currently being displayed, tap right below the title of the task (where the second task would be, as if the layout were not there).
7. As you will see, the text of the edition layout has changed, and it now displays the contents of the 2nd task. The same thing can be done for the 3rd task and so on. 
![edition layout with 2nd tasks contents](https://user-images.githubusercontent.com/7356565/117506571-15bb3180-af86-11eb-9719-8f7e476f9d12.png)

# Expected behaviour

The task update interface should have remained with task 1's contents, and should block any clicks over underlying elements.

YagoGG:
Fixed via #242.

PR #248, opened 2021-05-13
add Tags feature

lgraziano:
Adds the tags feature.
For now we can display and add tags by entering a todo-list and then pressing the button in the upper right corner.


app/src/androidTest/java/com/github/steroidteam/todolist/TagTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+        Bundle bundle = new Bundle();
+        bundle.putSerializable("list_id", UUID.randomUUID());
+        scenario =
+                FragmentScenario.launchInContainer(
+                        ItemViewFragment.class, bundle, R.style.Theme_Asteroid);
+    }
+
+    @Test
+    public void TagButtonOpensTagWindow() {
```
YagoGG, line 81: Methods should start with a lowercase letter.

app/src/androidTest/java/com/github/steroidteam/todolist/TagTest.java
```diff
@@ -0,0 +1,166 @@
...
+        onView(withId(R.id.itemview_tag_button)).perform(click());
+        onView(withId(R.id.layout_update_tags)).check(matches(isDisplayed()));
+    }
+
+    @Test
+    public void CreateTagButtonWorks() {
```
YagoGG, line 87: Methods should start with a lowercase letter.

app/src/main/java/com/github/steroidteam/todolist/model/todo/Tag.java
```diff
@@ -0,0 +1,76 @@
...
+        this.uuid = UUID.randomUUID();
+        this.body = body;
+        this.color = color;
+        listsIds = new HashSet<>();
+        listsIds.add(listId);
+        tags.add(this);
+    }
+
+    public Tag(String body, UUID listId) {
+        this(body, Color.WHITE, listId);
```
YagoGG, line 31: It'd be cool if the color were randomly generated, but this is just nitpicking. We can add the colour picker or whatever view we use for colours.

app/src/main/java/com/github/steroidteam/todolist/view/ItemViewFragment.java
```diff
@@ -202,6 +213,62 @@ public void checkBoxTaskListener(final int position, final boolean isChecked) {
...
+    public void tagSaveButton(View view) {
+        ConstraintLayout tagLayout = getView().findViewById(R.id.layout_update_tags);
+        tagLayout.setVisibility(View.GONE);
+    }
+
+    private void createTagButton(Tag tag, LinearLayout row) {
+        Button tagButton =
+                new Button(new ContextThemeWrapper(getContext(), R.style.TagInList), null, 0);
+        tagButton.setText(tag.getBody());
+        // tagView.setBackgroundColor(tag.getColor());
```
YagoGG, line 239: Should this be uncommented?

app/src/main/java/com/github/steroidteam/todolist/model/todo/Tag.java
YagoGG: In fact, do we even use this constructor?

YagoGG (commented):
Hey @lgraziano! Super cool work.

I have been able to test the feature and it works perfectly. The only issue I found is that the background of the layout is "transparent" to clicks, meaning that you can click the tasks underneath without noticing (see #237). However, it is not a big problem.

The code looks great, but I have some doubts about the object model. I see that labels contain lists within them, but I wonder if it shouldn't be all the way around: each list contains a "tags" attribute with a list of references to the corresponding label.

I feel like this would make much easier implementing the database backend later on, but I am not 100% sure, so I prefer that we discuss this before aproving/requesting changes.

Thanks!

theo-damiani:
> 
> 
> Hey @lgraziano! Super cool work.
> 
> I have been able to test the feature and it works perfectly. The only issue I found is that the background of the layout is "transparent" to clicks, meaning that you can click the tasks underneath without noticing (see #237). However, it is not a big problem.
> 
> The code looks great, but I have some doubts about the object model. I see that labels contain lists within them, but I wonder if it shouldn't be all the way around: each list contains a "tags" attribute with a list of references to the corresponding label.
> 
> I feel like this would make much easier implementing the database backend later on, but I am not 100% sure, so I prefer that we discuss this before aproving/requesting changes.
> 
> Thanks!

Hey @YagoGG, we discussed with Leandro, and I was telling him that maybe it was better to have a label model. 
So each label has a list of associated lists or notes. And when you add a tag to a list you just, call the database and add a UUID within the label. So you recover only the tag model which is really small: a list of UUID, maybe a text body, and one color.
If you do the other way around, each time you add a label to a list you will call the database to recover the entire todo list, and all this stuff just to add a UUID to the list. So I thought it was less consuming to do like Leandro decided.

And so you can have a folder /tags/ in the database as we did for voice memos or images. Also if you remove or add a new tag you just put or remove a file in the folder tags and you do not need to go all todolist to check: oh this one has the tag we need to remove it.

But for sure it is more difficult to print all the labels that are associated with one todolist.
What do you think?

app/src/main/java/com/github/steroidteam/todolist/model/TodoRepository.java
```diff
@@ -77,4 +83,27 @@ public void setTaskDueDate(int index, Date dueDate) {
                 .thenAccept(this.oneTodoList::setValue);
         this.database.getTodoList(todoListID).thenAccept(this.oneTodoList::setValue);
     }
+
+    private void setTagsList() {
+        List<UUID> tagsIds = oneTodoList.getValue().getTagsIds();
+        System.out.println("Tags IDs : " + tagsIds);
```
sydneyhauke, line 89: Why printlns here ?

sydneyhauke (changes requested):

app/src/main/java/com/github/steroidteam/todolist/model/TodoRepository.java
lgraziano: That was for fixing a bug, I'll remove it when the PR's ready for reviews

app/FAKE-LOCAL-PATH/user-data/fake id/test-data/data2
```diff
@@ -0,0 +1 @@
+Some data2 to write!
```
theo-damiani, line 1: remove this file

app/FAKE-LOCAL-PATH/user-data/fake id/test-data/data1
```diff
@@ -0,0 +1 @@
+Some data to write!
```
theo-damiani, line 1: remove this file

app/src/main/java/com/github/steroidteam/todolist/view/ItemViewFragment.java
```diff
@@ -305,6 +321,77 @@ public void checkBoxTaskListener(final int position, final boolean isChecked) {
...
+        DialogFragment newFragment =
+                new InputDialogFragment().newInstance(dialogListener, R.string.add_tag_suggestion);
+        newFragment.show(getParentFragmentManager(), "add_tag_dialog");
+    }
+
+    public void destroyTag(Tag tag) {
+        // TODO : Uncomment after fixing method in repository/database
+        // itemViewModel.destroyTag(tag);
+    }
+
```
theo-damiani, line 394: I tested and you can't delete a tag, I think it might come from this commented code

theo-damiani (approved):
Great work!

Your PR change a lot of things.
I tested it and it works well but you can't delete a tag.
Also, the tags are saved only locally and note in the remote database.

sydneyhauke (approved):

PR #262, opened 2021-05-18
Feature/note persistence

theo-damiani:


app/src/main/java/com/github/steroidteam/todolist/database/FirebaseDatabase.java
```diff
@@ -315,4 +316,60 @@ public FirebaseDatabase(@NonNull FirebaseFileStorageService storageService) {
...
+                        note -> {
+                            note.setHeader(headerID);
+                            return uploadNote(note);
+                        })
+                .thenCompose(note -> headerUploadFuture)
+                .thenApply(str -> null);
+    }
+
+    @Override
+    public CompletableFuture<Void> removeHeader(UUID noteID) {
```
sydneyhauke, line 348: Are you able to factorize this code with setHeaderNote ?

shoutizix (approved):
Hey !
I've just tested it and it works like a charm !
I took a picture and select it as header and even when restarting the app it was still correctly displayed.

The only little thing is that now our coverage will decrease a bit but if you can't manage to test it more well I think it's still ok to merge this because it's important to achieve persistency in all the app I think

PR #263, opened 2021-05-20
fix: Delete UserFactory user upon logout

YagoGG:
This was causing that the LoginActivity was not displayed after logout, and thus unauthorized requests were being sent to Firebase, causing errors.


shoutizix (approved):
Very very very nice line added

lgraziano (approved):
10/10

PR #264, opened 2021-05-20
Group Tasks by date

shoutizix:


shoutizix:
It works great now ! but I have to fix removeAllTask button which doesn't work anymore

theo-damiani (changes requested):
Hey @shoutizix,
Great job!

The code is very clean, but I run into somes issues while testing the feature:

- There is a blue border for each list of tasks: 


- In the picture above, the task finish the demo was in fact for the past week, so maybe you should create a category "past" or something to indicate that the due date has expired? I don't know what to do with the past dates, it's just a proposition, do what you prefer!

- Maybe it would be great to not displayed the date when there is none:
I mean if there is a date for today print "today" otherwise skip the title "today". And this for each category

Overall great job, I know nested recycler view can be tricky!
and nice coverage!

shoutizix:
Hey @theo-damiani !
Thanks a lot for the review ! I've made the changes you've requested, but on the point to not display the date I don't know what's the best because if I make them invisible then it just display blank between categories like this : 
Screenshot_20210525-121340

And I can't reorganize the categories because the task and all of the adapter are based on the fact that we know the position of each category and it won't change.

Maybe we could do that but I think this should be in an other PR.

Please let me know your opinion !

theo-damiani:
> 
> 
> Hey @theo-damiani !
> Thanks a lot for the review ! I've made the changes you've requested, but on the point to not display the date I don't know what's the best because if I make them invisible then it just display blank between categories like this :
> Screenshot_20210525-121340
> 
> And I can't reorganize the categories because the task and all of the adapter are based on the fact that we know the position of each category and it won't change.
> 
> Maybe we could do that but I think this should be in an other PR.
> 
> Please let me know your opinion !

Maybe if you make them disappear with setVisibility(Gone) and note Invisible, maybe the blank space will be gone

shoutizix:
> > Hey @theo-damiani !
> > Thanks a lot for the review ! I've made the changes you've requested, but on the point to not display the date I don't know what's the best because if I make them invisible then it just display blank between categories like this :
> > And I can't reorganize the categories because the task and all of the adapter are based on the fact that we know the position of each category and it won't change.
> > Maybe we could do that but I think this should be in an other PR.
> > Please let me know your opinion !
> 
> Maybe if you make them disappear with setVisibility(Gone) and note Invisible, maybe the blank space will be gone.
> Yes don't change all the things, it is very good but maybe just changing the visibility of the specific recycler view for a given date where there are no tasks will work.
> 
> And if you could check the padding between tasks it would be great. For example, in the above picture, your tasks don't seem to be centered.

Oh you're right for the padding I've just fix this. But I tried with the view gone already and it doesn't change anything.

shoutizix:
@theo-damiani 
Now it should be working !

theo-damiani (approved):
Tested again and,
man, it looks awesome, thanks for the changes!

Just to point some details even if I think the PR is ready to be merged:

For Tuesday this is just due to the time zone I think, so no problem.

But:
- You can see that Monday is in this week and not in Later.
- and when I put Thursday the task simply didn't appear.

shoutizix:
> Tested again and,
> man, it looks awesome, thanks for the changes!
> 
> Just to point some details even if I think the PR is ready to be merged:
> 
> For Tuesday this is just due to the time zone I think, so no problem.
> 
> But:
> 
> * You can see that Monday is in this week and not in Later.
> * and when I put Thursday the task simply didn't appear.

Thanks again for the feedback !
In fact it's normal, because the "This week" is for the next 7 days, so the next 5 days after tomorrow, and I've renamed it accordingly since it can lead to confusion.

VPellenc (approved):
The feature looks very good and the code is very clean! Thanks for the changes

PR #265, opened 2021-05-20
FIX/LiveData not updated after sync database

theo-damiani:


YagoGG (approved):
Looks good to me as well! Thanks @theo-damiani :)

PR #266, opened 2021-05-20
Sort the task by their date

shoutizix:
This is a part of PR #264 , so even if it is small, it could be merged before the end of the sprint

### DIFF
--- a/app/src/main/java/com/github/steroidteam/todolist/viewmodel/TodoListViewModel.java
+++ b/app/src/main/java/com/github/steroidteam/todolist/viewmodel/TodoListViewModel.java
@@ -1,6 +1,5 @@
 package com.github.steroidteam.todolist.viewmodel;
 
-import android.util.Log;
 import androidx.annotation.NonNull;
 import androidx.lifecycle.LiveData;
 import androidx.lifecycle.ViewModel;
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/github/steroidteam/todolist/viewmodel/TodoListViewModel.java
```diff
@@ -74,6 +75,7 @@ public void renameTask(int index, String newBody) {
     }
 
     public void setTaskDone(int index, boolean isDone) {
+        Log.d("DEBUG", todoListSelected.getValue().toString());
```
YagoGG, line 78: I'm guessing we don't want this here? ;)

YagoGG (approved):
Just left a really minor comment, but the rest looks great. Thanks a bunch, @shoutizix!

app/src/main/java/com/github/steroidteam/todolist/viewmodel/TodoListViewModel.java
shoutizix: Oh yep this one was forgotten thanks ^^

theo-damiani (approved):

PR #281, opened 2021-05-27
Done category in task layout

theo-damiani:
Thanks, Noah for the help.

I used his PR #264.

### DIFF
--- a/app/src/main/java/com/github/steroidteam/todolist/model/todo/TodoList.java
+++ b/app/src/main/java/com/github/steroidteam/todolist/model/todo/TodoList.java
@@ -112,6 +112,15 @@ public class TodoList {
         return this;
     }
 
+    public TodoList sortByDone() {
+        Collections.sort(
+                list,
+                (task, t1) -> {
+                    return Boolean.compare(task.isDone(), t1.isDone());
+                });
+        return this;
+    }
+
     @Override
     public String toString() {
         StringBuilder str = new StringBuilder();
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/github/steroidteam/todolist/model/todo/TodoList.java
```diff
@@ -112,6 +112,15 @@ public TodoList sortByDate() {
         return this;
     }
 
+    public TodoList sortByDone() {
+        Collections.sort(
+                list,
+                (task, t1) -> {
+                    return Boolean.compare(task.isDone(), t1.isDone());
```
sydneyhauke, line 119: I would like to know if this will change the order of the tasks in which they were added (I mean between non-done tasks, and otherwise between done tasks)  inside of a todolist and if yes, if it is acceptable or not.

theo-damiani: I did like Noah, so yes it changes the order of the task. But I think it is well handled in the ParentAdapter

app/src/main/java/com/github/steroidteam/todolist/view/adapter/ParentTaskAdapter.java
```diff
@@ -41,6 +42,7 @@ public ParentTaskAdapter(TodoAdapter.TaskCustomListener listener) {
         dateCategoryList.add("Next 5 days after Tomorrow");
         dateCategoryList.add("Later");
         dateCategoryList.add("Unknown");
+        dateCategoryList.add("Done");
```
lgraziano, line 45: I think it could be a good thing to move these strings to strings.xml

theo-damiani: Sure you right!

lgraziano (approved):
The feature works as intended and the code is clean 👌

shoutizix (approved):
We've seen it together and yes it works well, I haven't found any trouble by using this and the database still seems correct so I think we can merge it.

Nice !

PR #284, opened 2021-05-27
Increases test coverage of NoteDisplayFragment

sydneyhauke:
Title is self explanatory. I added a few more tests (those test if the note header image has been updated or not after selecting an image)


app/src/androidTest/java/com/github/steroidteam/todolist/NoteDisplayFragmentTest.java
```diff
@@ -186,4 +197,91 @@ public void addImageDialogWorks() {
...
     }
+
+    @Test
+    public void headerImageIsNotSetWorks() {
+        Intents.init();
+
+        onView(withId(R.id.note_header))
+                .check(
+                        matches(
+                                new BoundedMatcher<View, ConstraintLayout>(ConstraintLayout.class) {
```
YagoGG, line 208: Can you please move this to the `CustomMatchers` class?

YagoGG (approved):
Hey, @sydneyhauke! Thanks a lot for the changes.

I just left a minor comment regarding code structure.

As for the "haunted test", should we keep it or can we remove it? A third option would be to decorate it with `@Ignore`, but that would only make sense if we plan to use the test in the future.

Cheers!

theo-damiani (approved):

PR #310, opened 2021-05-28
Fix the notification based on date reminder

shoutizix:
I assigned an unique Id for pendingIntent and get the intent not directly in the constructor of the notification and it fixes the bug for me.

Please tell me if it works also on your device


theo-damiani (approved):
Maybe the static variable is not the most beautiful way to do it,
But I tested and it worked well so good job!

app/src/main/java/com/github/steroidteam/todolist/broadcast/ReminderDateBroadcast.java
```diff
@@ -19,16 +19,17 @@
     public static String REMINDER_DATE_CHANNEL_ID = "notifyDateChannelID";
     public static CharSequence NAME_DATE_CHANNEL = "ReminderDateChannel";
     public static String DESCRIPTION_DATE_CHANNEL = "Channel for the date reminder";
+    private static int unique_ID = 0;
```
sydneyhauke, line 22: It seems a bit hacky. It works though so we can keep it as is.

sydneyhauke (approved):

PR #313, opened 2021-06-01
fix: Do not display due dates when null

### LINKED ISSUE #289: Due date is displayed even if none
- create a todo list
- add a task with a due date
- add a second task without a due date
- open the "update layout" for the first task, you can see the due date
- open the second task and you will see the same due date

In fact, the update layout keeps the last due date. 
So if you create a third task with another task, you open it. Then if you open a task without a date you will see the due date from the third one. But if you open the first and then the second, you will see the first due date.

YagoGG:
Fixes #289.


theo-damiani:
Tested and indeed the bug is fixed!
Good job!

theo-damiani (approved):
Tested and indeed the bug is fixed!
Good job!

shoutizix (approved):
Just tested on my phone and works perfectly. Nice fix !

YagoGG:
@theo-damiani @shoutizix thanks a bunch for the review, merging! :)

PR #315, opened 2021-06-02
add display option for tags

lgraziano:
The tag view now has two rows :
1) The tags associated to the current list
2) The other existing tags

Unfortunately, the app sometimes crashes when doing several actions one after the other.


app/src/main/java/com/github/steroidteam/todolist/database/TodoListRepository.java
```diff
@@ -221,37 +217,98 @@ public void updateTask(UUID todoListID, int index, Task updatedTask) {
...
+                        str -> {
+                            List<Tag> unlinked = globalTags.getValue();
+                            unlinked.removeAll(localTags.getValue());
+                            unlinkedTags.postValue(unlinked);
+                        });
+
+        remoteDatabase.putTag(tag);
+    }
+
+    public void removeTagFromTodolist(UUID todoListID, UUID tagId) {
```
sydneyhauke, line 266: I wonder if we are not starting to pile-up a lot of round-trips when dealing with the database.

app/src/main/java/com/github/steroidteam/todolist/view/misc/TagView.java
```diff
@@ -14,48 +14,109 @@
...
+        this.viewModel = viewModel;
+        tagLayout = root.findViewById(R.id.layout_update_tags);
+        localRow = root.findViewById(R.id.tag_row_local);
+        unlinkedRow = root.findViewById(R.id.tag_row_global);
+        init();
+    }
+
+    private void init() {
+        localTags = viewModel.getTagsFromList();
+        if (localTags == null) {
```
sydneyhauke, line 41: Why do null checks and not init once and for-all in the constructor ?

YagoGG (approved):
Reviewed over Discord with Leandro. Great job!

shoutizix (approved):
Reviewed and tested with Leandro on Discord. Nice job

PR #321, opened 2021-06-04
fix: Solve dark mode-related UI problems

YagoGG:


sydneyhauke (approved):

shoutizix (approved):
Nice, I solve all the issues that we found with the dark-theme

PR #322, opened 2021-06-04
fix sync method

theo-damiani:


shoutizix (approved):
Perfect

VPellenc (approved):
Thanks for the fix

PR #327, opened 2021-06-07
bug-fix/tagsPersistence

lgraziano:
Fixes the persistance issue for the tags.


lgraziano:
There is still a remaining issue : sometimes you have to press the button several times before the global tags (the ones existing but not linked to the current todolist) get displayed

theo-damiani (approved):

